PR #2, opened 2023-01-13
docs: fix example.tf 

tekumara:
Running `terraform plan` on the example when dev overrides are enabled failed because it was using the aws provider without a dev override:
```
❯ terraform plan
╷
│ Warning: Provider development overrides are in effect
│ 
│ The following provider development overrides are set in the CLI configuration:
│  - prefecthq/prefect in /Users/tekumara/go/bin
│ 
│ The behavior may therefore not match any released version of the provider and applying changes may cause the state
│ to become incompatible with published releases.
╵
╷
│ Error: Inconsistent dependency lock file
│ 
│ The following dependency selections recorded in the lock file are inconsistent with the current configuration:
│   - provider registry.terraform.io/hashicorp/aws: required by this configuration but no version is selected
│ 
│ To make the initial dependency selections that will initialize the dependency lock file, run:
│   terraform init
```

I've commented out the aws section of the example so this now works, and fixed a few other issues that prevented the example from running successfully.

### DIFF
--- a/examples/main.tf
+++ b/examples/main.tf
@@ -31,7 +31,7 @@ resource "prefect_service_account" "test" {
   ]
 }
 
-# example showing how to inject the store the api key as a secret value
+# example showing how to store the api key in an AWS Secrets Manager secret
 
 # resource "aws_secretsmanager_secret" "this" {
 #   name = "prefect-svc.apikey"
--- a/examples/main.tf
+++ b/examples/main.tf
@@ -1,7 +1,7 @@
 terraform {
   required_providers {
     prefect = {
-      source = "registry.terraform.io/PrefectHQ/prefect"
+      source = "PrefectHQ/prefect"
     }
   }
   required_version = ">= 1.0"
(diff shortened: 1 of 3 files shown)


examples/main.tf
```diff
@@ -22,20 +22,22 @@ resource "prefect_service_account" "test" {
...
-resource "aws_secretsmanager_secret" "this" {
-  name = "prefect-svc.apikey"
-}
+# example showing how to store the api key in an AWS Secrets Manager secret
 
-resource "aws_secretsmanager_secret_version" "this" {
-  secret_id     = aws_secretsmanager_secret.this.id
-  secret_string = prefect_service_account.test.api_keys[0].key
-}
+# resource "aws_secretsmanager_secret" "this" {
```
gabcoyne, line 36: Do you want this commented out? I don't see an issue with it uncommented as it's an example

tekumara: Yeh I had to comment this out otherwise when dev overrides are enabled `terraform plan` fails (see the error in the PR description).

I think the common practice here is to have minimal examples that only use a single terraform provider .. for more info see https://github.com/hashicorp/terraform/issues/27459#issuecomment-1382126220.

jamiezieziula (approved):

PR #8, opened 2023-03-14
Initial implementation of Prefect 2 provider

### LINKED ISSUE #3: Support for Prefect Cloud v2
Hello!

I'd love to include some Terraform code to create Infrastructure and Storage blocks in my Prefect v2 Cloud so both AWS cloud infrastructure and Prefect components remain tightly coupled. 

Are there plans to update this provider and cover your cloud v2 APIs?

maikelpenz:
This is a pull request to put the basics of terraform automation for Prefect 2.0 in place.

The content of this is an exact copy of an implementation [I came up with recently](https://github.com/maikelpenz/terraform-provider-prefect2) and my thoughts about what the 2.0 needs can be found on its README file (also pushed as part of this PR). Happy to get more eyes and hands on this for sure.

Something to note is that I created a branch on my forked repo named "2.x", keeping the existing 1.0 implementation as it is in the main branch, similar to what has been done for the Prefect library. If we think we'd like to proceed with a similar approach it would be better to get this merged into "2.x" instead of "main".

(Actually we shouldn't merge this as it is because it will override the 1.0 implementation on main, but we should decide first where 2.0 should live)

Resolves: #3


cforcey:
This is great to see @maikelpenz!  We have transitioned from Prefect 1 to Prefect 2 and will feel much more comfortable when our Prefect 2.0 settings can join the rest of our infrastructure in Terraform. Thank you and another vote for making a quick decision on the new branch and letting this out in the wild, even as a sub-1.0 release for testing.

LICENSE
```diff
@@ -1,4 +1,3 @@
-
                                  Apache License
```
jawnsy, line 1: Can you please revert this change? `git checkout main LICENSE` should do the trick

go.mod
```diff
@@ -1,74 +1,48 @@
-module terraform-provider-prefect
+module terraform-provider-prefect2
```
jawnsy, line 1: Go modules use the repository path (so our previous module name was incorrect), can you change this `github.com/PrefectHQ/terraform-provider-prefect` instead?

prefect2/data_source_block_documents.go
```diff
@@ -0,0 +1,138 @@
+package prefect2
```
jawnsy, line 1: We never published a `prefect` provider on the Terraform Registry, so I'd propose that we simply call this `prefect`

prefect2/data_source_work_queues.go
```diff
@@ -0,0 +1,140 @@
...
+						},
+					},
+				},
+			},
+		},
+	}
+}
+
+func dataSourceWorkQueuesRead(ctx context.Context, d *schema.ResourceData, m interface{}) diag.Diagnostics {
+	c := m.(*hc.Client)
```
jawnsy, line 83: I'm not too familiar with the Terraform plugin infrastructure, is it possible that they pass us a different parameter? Do we need to check this type assertion to prevent a panic, or can we expect that `m` is always a Client?

prefect2/data_source_work_queues.go
```diff
@@ -0,0 +1,140 @@
...
+
+	var workQueuesOutput []interface{}
+
+	if workQueueId != "" {
+		workQueue, err := c.GetWorkQueue(workQueueId, workspaceId)
+		if err != nil {
+			return diag.FromErr(err)
+		}
+
+		workQueues := make([]hc.WorkQueue, 1, 1)
```
jawnsy, line 99: small nitpick (feel free to ignore, we can clean up later):

Manually specifying the length and capacity of array slices is not common in Go, except for cases requiring performance optimization by preallocating memory. Instead, the two preferred ways to do this in Go are to use `append`:

```go
	workQueues := []WorkQueue{}
	workQueues = append(workQueues, WorkQueue{
		Id:      "hello",
		Created: "world",
	})
```

or to define it as a literal, which I think is preferred in this case:

```go
	workQueues = []WorkQueue{
		{
			Id:      "hello",
			Created: "world",
		},
	}
```

You can see both examples here: https://go.dev/play/p/8eaiJ4PCk1J

I haven't tried it, but in this case, I think this should work:

```suggestion
		workQueues := []hc.WorkQueue{
			*workQueue,
		},
```

prefect2_api/block_document.go
```diff
@@ -0,0 +1,121 @@
...
+
+	err = json.Unmarshal(body, &blockDocuments)
+	if err != nil {
+		return nil, err
+	}
+
+	return blockDocuments, nil
+}
+
+func (c *Client) GetBlockDocument(blockDocumentId string, workspaceID string) (*BlockDocument, error) {
```
jawnsy, line 32: Recent versions of Go now pass `ctx.Context` parameter as the first parameter by convention, because they are required to support timeouts and request cancellation. We can take care of it after merging, so it's not a big deal, but I think this function would be more idiomatic if they looked like this:

```suggestion
func (c *Client) GetBlockDocument(ctx context.Context, blockDocumentId string, workspaceID string) (*BlockDocument, error) {
	req, err := http.NewRequestWithContext(ctx, "GET", fmt.Sprintf("%s/accounts/%s/workspaces/%s/block_documents/%s", c.PrefectApiUrl, c.PrefectAccountId, workspaceID, blockDocumentId), nil)
```

See: https://pkg.go.dev/net/http#NewRequestWithContext and https://go.dev/blog/context

prefect2_api/block_document.go
```diff
@@ -0,0 +1,121 @@
...
+	req, err := http.NewRequest("GET", fmt.Sprintf("%s/accounts/%s/workspaces/%s/block_documents/%s", c.PrefectApiUrl, c.PrefectAccountId, workspaceID, blockDocumentId), nil)
```
jawnsy, line 33: We can refactor later, but a lot of our client code in Python includes the `accounts` and `workspaces` path parts as part of the URL field, so that we can support both Prefect Server and Prefect Cloud, so that may be something to consider here as well.

For example, rather than defining things this way, we could define `c.PrefectApiUrl` as `https://api.prefect.cloud/api/accounts/741fcb04-d564-11ed-afa1-0242ac120002/workspaces/7b916690-d564-11ed-afa1-0242ac120002` and then the code would only need to append `block_documents`. We would need other changes as well to better support Server (we would want to restrict resources that are not currently compatible with Server)

 Feel free to leave this as-is and we can log an issue to revise later.

examples/data_sources.tf
```diff
@@ -0,0 +1,43 @@
...
+
+# S3 block type and schema (helpers)
+
+data "prefect2_block_types" "s3_block_type" {
+    workspace_id = prefect2_workspace.terraform-workspace.id
+    slug = "s3"
+}
+
+data "prefect2_block_schemas" "s3_block_schema" {
+    workspace_id = prefect2_workspace.terraform-workspace.id
```
jawnsy, line 29: Can you run this through `terraform fmt`?

.DS_Store
jawnsy, line 1: Please remove this file (`git rm .DS_Store` should do the trick)

prefect2_api/workspace.go
```diff
@@ -0,0 +1,75 @@
...
+	err = json.Unmarshal(body, &workspace)
+	if err != nil {
+		return nil, err
+	}
+
+	return &workspace, nil
+}
+
+func (c *Client) CreateWorkspace(workspace Workspace) (*Workspace, error) {
+	rb, err := json.Marshal(workspace)
```
jawnsy, line 53: suggestion: You can use `json.NewEncoder()` to create a stream and pass that to `NewRequest` instead. This has better performance because it avoids memory copies.

jawnsy (commented):
Hey there! Thanks so much for your contribution, this looks like an awesome start at something we've wanted to tackle for awhile. I left a few comments throughout for your consideration, but most of them are minor and we can tackle them later.

We should be able to accept this change in the next few days, but it may be a few weeks until we can do a deeper review and get something published on the Terraform registry.

maikelpenz:
> Hey there! Thanks so much for your contribution, this looks like an awesome start at something we've wanted to tackle for awhile. I left a few comments throughout for your consideration, but most of them are minor and we can tackle them later.
> 
> We should be able to accept this change in the next few days, but it may be a few weeks until we can do a deeper review and get something published on the Terraform registry.

Thank you very much for jumping into this PR @jawnsy. Give me some days to go through your review 👍.

prefect2/data_source_block_documents.go
maikelpenz: @jawnsy for people using this automation for Prefect 1, wouldn't changing this cause problems?

prefect2/data_source_work_queues.go
maikelpenz: I put a helper function now that is uses for all these methods. The helper function `getClient` checks for the type.

prefect2/data_source_work_queues.go
maikelpenz: Thanks for the suggestion. This has been changed.

prefect2/data_source_block_documents.go
jawnsy: No, we're not planning to publish a Terraform provider for Cloud 1.0, so yes, I think just `prefect` makes sense!

prefect2_api/block_document.go
maikelpenz: Awesome, I've updated all functions inside the `_api` folder

prefect2_api/block_document.go
maikelpenz: Hey @jawnsy I could refactor this one but I'd rather leave this with your team. You will know more about the differences and similarities between `Prefect Server` and `Prefect Cloud`.

One thing to note is that `workspaces` are created as part of the automation, so while we can set the terraform variable  up to the "account" level we'd still need to make it work to append the workspace in question before calling the `prefect_api` method. Think on a case where the automation sets 3 different workspaces, would you update the client on the fly or pass the URL to every function in the `prefect_api` side and not have it set in the client..

I hope it makes sense.

prefect2_api/workspace.go
maikelpenz: Done!

go.mod
maikelpenz: hmm I'm not sure how we get this to work. When I change this to `github.com/PrefectHQ/terraform-provider-prefect` I get first an error where I reference the module, which is fine, but then when I update it I get this second error complaining about `import cycle not allowed`.

Do you have any thoughts @jawnsy ?

image

jawnsy: No worries! Leave this as it is, I'll take a look after we merge the initial version

prefect2_api/block_document.go
jawnsy: Yup, that makes sense! Thank you :)

maikelpenz:
@jawnsy most of your comments have been addressed. Two things to note:

* To create "blocks" through terraform I need to retrieve the "block schema" for what I want (e.g S3, Kubernetes). And the API I am using is the one that gets the [block schema by checksum](https://app.prefect.cloud/api/docs#tag/Block-schemas/operation/read_block_schema_by_checksum_api_accounts__account_id__workspaces__workspace_id__block_schemas_checksum__checksum__get). This is because I found out the same checksum is used across workspaces. However, I am not sure this is the same checksum "across accounts". This is not a big problem as every Prefect Cloud user can adjust to the checksum set for their account, but if there was an API available to get block schema by name it would be simpler (and potentially global).

* Second note is around merging this to master. I understand you'd be happy to get this as it is into master but I just want to make sure @tekumara is across this as he released the 1.0 version of this automation. If we want to keep both maybe we put this in a separate branch. Will leave this with you and I can make changes to my PR if you wish.

jawnsy (approved):

cforcey:
Thanks so much everyone for this effort!  We can put terraforming our Prefect 2 installation on the roadmap thanks to this nice surprise. Best wishes!

tekumara:
Thanks for checking @maikelpenz ! No problem for us, we have an internal version of the provider so we're happy for this to be a Prefect 2 only provider.

PR #17, opened 2023-05-19
chore: use Go module conventions

jawnsy:
By convention, Go modules use the full GitHub repository path when they are hosted on GitHub. This allows "go get" to identify and load modules from GitHub directly.


jamiezieziula (approved):
🚀

PR #19, opened 2023-05-23
chore: add Dependabot settings

jawnsy:


jamiezieziula (approved):
go depbot! just removed quotes but other than that looks good

PR #22, opened 2023-05-24
feat: migrate to Terraform Plugin Framework

jawnsy:
This change migrates to the new Terraform Plugin Framework, which removes compatibility for old (pre-1.0) versions of Terraform.


jamiezieziula (approved):
wow 🚀

PR #23, opened 2023-05-26
feat: implement variable data source and resource

jawnsy:
* Add tfstate backup to ignore
* Customize rules for gocritic
* Implement VariableClient to work with variables
* Implement data source for retrieving variables by ID and Name
* Implement resource for creating/updating variables

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 NAME=prefect
 BINARY=terraform-provider-${NAME}
 
-default: build test
+default: build
 .PHONY: default
 
 help:
@@ -10,6 +10,7 @@ help:
 	@echo "This project defines the following build targets:"
 	@echo ""
 	@echo "  build - compiles source code"
+	@echo "  lint - run static code analysis"
 	@echo "  test - run automated tests"
 	@echo "  clean - removes built artifacts"
 .PHONY: help
@@ -25,6 +26,10 @@ $(BINARY):
 clean:
 	rm -vrf build/
 .PHONY: clean
+
+lint:
+	golangci-lint run
+.PHONY: lint
 
 test:
 	gotestsum --max-fails=10 ./...
--- a/Readme.md
+++ b/Readme.md
@@ -2,19 +2,16 @@
 
 Terraform provider for [Prefect 2](https://github.com/PrefectHQ/prefect) and [Prefect Cloud](https://app.prefect.cloud).
 
-## Features:
+## Resources
 
-* Data Sources:
-    - Workspaces
-    - Work Queues
-    - Block Types
-    - Block Schemas
-    - Block Documents
+* Variable (`prefect_variable`)
+* Work Pool (`prefect_work_pool`)
 
-* Resources:
-    - Workspace (Create)
-    - Work Queue (Create, Update, Delete)
-    - Block (Create, Update, Delete)
+## Data Sources
+
+* Variable (`prefect_variable`)
+* Work Pool (`prefect_work_pool`)
+* Work Pools (`prefect_work_pools`)
 
 ## Deployment:
 The "examples" folder makes use of this local provider.   
@@ -93,6 +90,7 @@ On MacOS, you can use `brew` to install necessary dependencies:
 
 ```shell
 brew install \
+  go \
   gotestsum \
   golangci-lint \
   goreleaser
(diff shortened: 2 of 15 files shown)


internal/api/client.go
```diff
@@ -1,5 +1,7 @@
 package api
 
+// PrefectClient returns clients for different aspects of our API.
```
parkedwards, line 3: nice

Makefile
```diff
@@ -26,6 +27,10 @@ clean:
 	rm -vrf build/
 .PHONY: clean
 
+lint:
+	golangci-lint run
```
parkedwards, line 31: prob a dumb q - would this catch anything that gofmt or the compile step wouldn't catch?

internal/client/variables.go
```diff
@@ -0,0 +1,166 @@
...
+		return nil, fmt.Errorf("failed to decode response: %w", err)
+	}
+
+	return &variable, nil
+}
+
+// List returns a list of variables matching filter criteria.
+func (c *VariablesClient) List(ctx context.Context, filter api.VariableFilter) ([]api.Variable, error) {
+	_ = ctx
+	_ = filter
```
parkedwards, line 68: could you help me understand what's going on here? so the `_` represents the `VariablesClient` interface itself?

Makefile
jawnsy: yes, golangci-lint collects hundreds of lint tools, the compiler does some checks but not style linting

https://golangci-lint.run/usage/linters/

internal/client/variables.go
parkedwards: ahhh got it

internal/client/variables.go
```diff
@@ -0,0 +1,166 @@
...
+	req, err := http.NewRequestWithContext(ctx, http.MethodDelete, c.endpoint+"/variables/"+variableID.String(), http.NoBody)
+	if err != nil {
+		return fmt.Errorf("error creating request: %w", err)
+	}
+
+	resp, err := doRequest(c.hc, c.apiKey, req)
+	if err != nil {
+		return fmt.Errorf("http error: %w", err)
+	}
+	defer resp.Body.Close()
```
parkedwards, line 159: we could probably wrap these in `doRequest()` and expose the method/path/params in the function sig.  not required though

internal/provider/datasources/work_pool.go
```diff
@@ -66,94 +68,102 @@ func (d *WorkPoolDataSource) Configure(_ context.Context, req datasource.Configu
...
-			},
-		},
+		Attributes: workPoolAttributes,
 	}
 }
 
 // Read refreshes the Terraform state with the latest data.
 func (d *WorkPoolDataSource) Read(ctx context.Context, req datasource.ReadRequest, resp *datasource.ReadResponse) {
-	var data WorkPoolSourceModel
+	var model WorkPoolDataSourceModel
```
parkedwards, line 123: nice, better

internal/client/variables.go
jawnsy: no, because defers run after the function returns, so that would close things early, e.g.

1. Call doRequest
2. defer resp.Body.Close()
3. Return from doRequest
4. Defer runs (body is closed)
5. We try to read from the body, but it's already closed

internal/provider/resources/work_pool.go
```diff
@@ -42,7 +44,9 @@ type WorkPoolResourceModel struct {
 	BaseJobTemplate  types.String `tfsdk:"base_job_template"`
 }
 
-//nolint:ireturn
+// New returns a new WorkPoolResource.
```
parkedwards, line 47:
```suggestion
// NewWorkPoolResource returns a new WorkPoolResource.
```

parkedwards (approved):
nice work!

Readme.md
```diff
@@ -2,19 +2,16 @@
...
+## Resources
 
-* Data Sources:
-    - Workspaces
-    - Work Queues
-    - Block Types
-    - Block Schemas
-    - Block Documents
+* Variable (`prefect_variable`)
+* Work Pool (`prefect_work_pool`)
```
jamiezieziula, line 8: im late to the party but this is awesome!

internal/provider/resources/work_pool.go
jawnsy: I missed this comment, will update in my next PR

PR #37, opened 2023-06-12
feat: add workspace support

### LINKED ISSUE #31: Resource and Data Source: prefect_workspace
Create a resource and data source for managing workspaces.

```terraform
resource "prefect_workspace" "foo" {
    name = "My test workspace"
}

data "prefect_workspace" "foo" {
    id = "6afbd2d8-48c5-4adf-969b-f2b42dd6acc8"
}
```

jawnsy:
* Allow provider to create and destroy workspaces in an account
* Add a default provider-scoped account_id and workspace_id for convenience
* Add utility functions to generate account- and workspace-scoped route prefixes, used when creating a client for account- or workspace-scoped resources

Closes: #31

### DIFF
--- a/internal/api/workspaces.go
+++ b/internal/api/workspaces.go
@@ -1,0 +1,41 @@
+package api
+
+import (
+	"context"
+
+	"github.com/google/uuid"
+)
+
+// WorkspacesClient is a client for working with workspaces.
+type WorkspacesClient interface {
+	Create(ctx context.Context, data WorkspaceCreate) (*Workspace, error)
+	Get(ctx context.Context, workspaceID uuid.UUID) (*Workspace, error)
+	Update(ctx context.Context, workspaceID uuid.UUID, data WorkspaceUpdate) error
+	Delete(ctx context.Context, workspaceID uuid.UUID) error
+}
+
+// Workspace is a representation of a workspace.
+type Workspace struct {
+	BaseModel
+	AccountID              uuid.UUID `json:"account_id"`
+	Name                   string    `json:"name"`
+	Description            *string   `json:"description"`
+	Handle                 string    `json:"handle"`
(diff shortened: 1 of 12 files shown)


internal/api/workspaces.go
```diff
@@ -0,0 +1,41 @@
...
+	Get(ctx context.Context, workspaceID uuid.UUID) (*Workspace, error)
+	Update(ctx context.Context, workspaceID uuid.UUID, data WorkspaceUpdate) error
+	Delete(ctx context.Context, workspaceID uuid.UUID) error
+}
+
+// Workspace is a representation of a workspace.
+type Workspace struct {
+	BaseModel
+	AccountID              uuid.UUID `json:"account_id"`
+	Name                   string    `json:"name"`
```
parkedwards, line 21: should the `Name` key be a pointer too?

internal/client/util.go
```diff
@@ -3,8 +3,45 @@ package client
...
 )
 
+// getAccountScopedURL constructs a URL for an account-scoped route
+func getAccountScopedURL(endpoint string, accountID uuid.UUID, route string) string {
+	var sb strings.Builder
+	sb.WriteString(endpoint)
+	sb.WriteString("/accounts/")
+	sb.WriteString(accountID.String())
+	sb.WriteRune('/')
+	sb.WriteString(route)
```
parkedwards, line 18: probably revealing my ignorance here, but why not just construct through `sprintf`?

parkedwards (approved):
nice

internal/client/util.go
jawnsy: No strong reason, really, just preference. I think this should perform marginally better than `fmt.Sprintf` because it avoids some necessary formatting of the format string and would support preallocating the buffer.

This approach is marginally beneficial for conditionals, since the alternative is to maintain two similar Sprintf statements. Arguably that would be more readable, I'm not opposed to refactoring this later 🤷

Issue #27, opened 2023-06-06
Resource and Data Source: prefect_service_account


jawnsy:
Create a resource and data source for managing service accounts.

```terraform
resource "prefect_service_account" "foo" {
}

data "prefect_service_account" "foo" {
}
```

PR #49, opened 2023-07-24
feat: Service Accounts Datasource + Resource

armalite:
Service Accounts tf resource/datasource implementation

This PR implements Prefect 2 'service accounts' (known as 'bots' in the API spec) in the terraform provider. It enables the creation of both service account resources and data sources.

The implementation is broken down akin to the other resource implementations: API request definitions, API client, Terraform resource and datasource implementation

This PR does not implement API key rotation for the service accounts, which will need to be implemented after.


parkedwards:
hey @armalite -- i'm part of the platform team at Prefect, where we've recently gotten some new internal targets on publishing this provider + have resumed concerted development efforts here. as I started work on #27, I just noticed your PR.  this is fantastic work and i don't want to step on any of it. 

Given some of these new internal targets, I'm tasked with pushing things over the finish line on a more accelerated timeline, so we can get this provider out to ya'll.  that said, I definitely don't want to apply any pressure on you since you've already generously given your free time and energy here -- wanted to see if you had any thoughts/preferences on tag-teaming this, or possibly handing it off to us?  I'd also totally understand + respect if you're committed to taking this home on your own, in which case I'd love to sync up on an ETA

armalite:
>  wanted to see if you had any thoughts/preferences on tag-teaming this, or possibly handing it off to us? I'd also totally understand + respect if you're committed to taking this home on your own, in which case I'd love to sync up on an ETA

Hi @parkedwards, great to hear this is being pushed for! :-) Out of curiosity what kind of timeline are you looking at? 

I'm quite tied up elsewhere at the moment but the plan was to wrap this up in the next couple of weeks - at least to a state where it can then be built further upon. If you need to get this through faster, I'd be happy to hand this over to you or tag-team and get this over the line (it's almost there I think). I've enabled maintainer edits, so would you be happy to continue working directly on this PR?

internal/provider/datasources/service_account.go
```diff
@@ -0,0 +1,177 @@
...
+// NewServiceAccountDataSource returns a new ServiceAccountDataSource.
+//
+//nolint:ireturn // required by Terraform API
+func NewServiceAccountDataSource() datasource.DataSource {
+	return &ServiceAccountDataSource{}
+}
+
+// Metadata returns the data source type name.
+func (d *ServiceAccountDataSource) Metadata(_ context.Context, req datasource.MetadataRequest, resp *datasource.MetadataResponse) {
+	resp.TypeName = req.ProviderTypeName + "_service_account"
```
parkedwards, line 46: on second pass, `service_account` is the correct nomenclature over `bot` 👍

internal/provider/datasources/service_account.go
```diff
@@ -0,0 +1,177 @@
...
+
+	d.client = client
+}
+
+var serviceAccountAttributes = map[string]schema.Attribute{
+	"id": schema.StringAttribute{
+		Computed:    true,
+		CustomType:  customtypes.UUIDType{},
+		Description: "Service Account UUID",
+		Optional:    true,
```
parkedwards, line 73: the `id` attribute may need to be `Required: true`

parkedwards:
> If you need to get this through faster, I'd be happy to hand this over to you or tag-team and get this over the line. If maintainer edits are enabled, would you be happy to continue working directly on this PR?

@armalite i think this sounds like the right way to go - let's do this. You're 99% of the way there on this PR, and I'm thinking we can rope this off to the core CRUD functionality.  I'll begin pushing a few commits here for you to take a look when you're able, with the aim of merging this PR by this week.  In a followup, I'll begin extending your work to include handling api key rotation updates for the service account resource

> It's almost there I believe - but may need some refactoring if we need to stick with the 'bot' over 'service account' naming

I left a comment above, but your intuition on the `service_account` nomenclature is better here, so I'm going to close out #63

parkedwards:
relates to https://github.com/PrefectHQ/terraform-provider-prefect/issues/27

Makefile
```diff
@@ -31,6 +31,9 @@ lint:
 	golangci-lint run
 .PHONY: lint
 
+install: clean build 
+	echo "@TODO Placeholder install - move built provider to ~.terraform.d/plugins/"
```
parkedwards, line 35: 👍 https://github.com/PrefectHQ/terraform-provider-prefect/issues/62

internal/provider/datasources/service_account_test.go
```diff
@@ -0,0 +1,45 @@
+package datasources_test
```
parkedwards, line 1: @armalite i added an acceptance test here against a dedicated testing environment, which verified your creation / deletion logic for the `prefect_service_account` resource as well as catching a few small bugs with the datasource 🙌

internal/provider/datasources/workspace_role_test.go
```diff
@@ -1,67 +1,84 @@
...
 
 	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
 	"github.com/prefecthq/terraform-provider-prefect/internal/testutils"
 )
 
-func TestWorkspaceRolesDataSource(t *testing.T) {
-	t.Parallel()
-	resource.Test(t, resource.TestCase{
+//nolint:paralleltest // we use the resource.ParallelTest helper instead
+func TestAccDatasource_workspace_role_defaults(t *testing.T) {
```
parkedwards, line 12: just refactoring a separate acceptance test, unrelated to service accounts

parkedwards (approved):

parkedwards:
awesome work on this!

Issue #40, opened 2023-06-12
Resource and Data Source: prefect_workspace_role


jawnsy:
Create a resource and data source for managing workspace-scoped roles:

```
resource "prefect_workspace_role" "foo" {
}

data "prefect_workspace_role" "foo" {
}
```

Issue #30, opened 2023-06-07
Automated tests for Terraform provider


jawnsy:
The Terraform Plugin Framework includes a test harness for writing integration tests: https://developer.hashicorp.com/terraform/plugin/framework/acctests

Our automated tests would require an external endpoint and secrets, though we can optionally mock out the remote server or use Prefect server for some tests. The Terraform test framework does not support unit testing, so validating errors or warnings may be more difficult (it is not currently possible to check for warnings given a Terraform configuration/run; see: https://github.com/hashicorp/terraform-plugin-testing/issues/69)

There is currently no test coverage for this provider and changes are tested manually on local development machines. It would be preferable to automate these tests, so that we can more easily review and accept pull request contributions.

parkedwards:
https://developer.hashicorp.com/terraform/plugin/testing/acceptance-tests

parkedwards:
also configure helpers for:
* precheck
* sweeps

PR #68, opened 2023-10-24
feat: add `workspace_role` datasource

parkedwards:
relates to https://github.com/PrefectHQ/terraform-provider-prefect/issues/40

this PR implements the `datasource` for the `workspace_role` resource

we also introduce an acceptance test, using the `terraform-plugin-testing` library


jamiezieziula (approved):

internal/provider/datasources/workspace_role_test.go
```diff
@@ -0,0 +1,67 @@
+package datasources_test
```
parkedwards, line 1: relates to https://github.com/PrefectHQ/terraform-provider-prefect/issues/30

PR #70, opened 2023-10-25
feat: add `workspace_role` resource

### LINKED ISSUE #40: Resource and Data Source: prefect_workspace_role

parkedwards:
resolves https://github.com/PrefectHQ/terraform-provider-prefect/issues/40


internal/api/workspace_role.go
```diff
@@ -19,7 +19,6 @@ type WorkspaceRole struct {
 	BaseModel
 	Name            string     `json:"name"`
 	Description     *string    `json:"description"`
-	Permissions     []string   `json:"permissions"`
```
parkedwards, line 22: `permissions` are a deprecated version of `scopes`, and can be removed here

internal/provider/resources/workspace_role.go
```diff
@@ -0,0 +1,375 @@
...
+	model.ID = types.StringValue(role.ID.String())
+	model.Created = customtypes.NewTimestampPointerValue(role.Created)
+	model.Updated = customtypes.NewTimestampPointerValue(role.Updated)
+
+	model.Name = types.StringValue(role.Name)
+	model.Description = types.StringPointerValue(role.Description)
+	model.AccountID = customtypes.NewUUIDPointerValue(role.AccountID)
+	model.InheritedRoleID = customtypes.NewUUIDPointerValue(role.InheritedRoleID)
+
+	// NOTE: here, we'll omit updating the TF state with the scopes returned from the API
```
parkedwards, line 135: let's say a practitioner configures a `manage_work_pools` scope in terraform.  Prefect Cloud will create the role + automatically add subordinate scopes, like `see_work_pools` and `write_work_pools`.  This will always cause a conflict on apply, as the practitioner will only have defined `manage_work_pools`

in this case, it's ok to hold onto the user-defined parameters, as they'll more like request inputs vs. a full declarative reflection of the scopes assigned to this role

internal/provider/resources/workspace_role_test.go
```diff
@@ -0,0 +1,63 @@
+package resources_test
```
parkedwards, line 1: just noting that these tests are not yet configured to run in GHA, as there's some extra setup we'll need to do around configuring the test provider to properly inherit env vars.  we'll also want to define a precheck function, which testcases can hook into to ensure all the necessary parameters exist

https://github.com/PrefectHQ/terraform-provider-prefect/issues/30

jimid27 (approved):
plz teach me Go one day

PR #73, opened 2023-10-26
feat: add ability to rotate Service Account API Keys

### LINKED ISSUE #27: Resource and Data Source: prefect_service_account

parkedwards:
resolves https://github.com/PrefectHQ/terraform-provider-prefect/issues/27

This PR extends #49, specifically on resource `Update()`.  Here, we'll conditionally call `POST /bots/rotate_api_key` if the key's expiration date field is changed, either between two different end dates OR transition to/from `null` (aka a key that never expires)

there were also some fixes done to the resource model, as well as some custom logic to read from State in order to keep the api key from changing without the correct trigger

finally, some acceptance tests and a key rotation example was added

### DIFF
--- a/examples/service_accounts.tf
+++ b/examples/service_accounts.tf
@@ -1,3 +1,9 @@
+provider "time" {}
+resource "time_rotating" "ninety_days" {
+    rotation_days = 90
+}
+
 resource "prefect_service_account" "sa_example" {
-    name = "beautiful-service-account-example"
+    name = "my-service-account"
+    api_key_expiration = time_rotating.ninety_days.rotation_rfc3339
 }
--- a/internal/provider/resources/service_account.go
+++ b/internal/provider/resources/service_account.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"time"
 
-	"github.com/hashicorp/terraform-plugin-framework/diag"
 	"github.com/hashicorp/terraform-plugin-framework/path"
 	"github.com/hashicorp/terraform-plugin-framework/resource"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
@@ -13,6 +12,7 @@ import (
 
 	"github.com/prefecthq/terraform-provider-prefect/internal/api"
 	"github.com/prefecthq/terraform-provider-prefect/internal/provider/customtypes"
+	"github.com/prefecthq/terraform-provider-prefect/internal/provider/helpers"
 )
 
 var (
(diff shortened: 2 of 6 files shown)


examples/service_accounts.tf
```diff
@@ -1,3 +1,9 @@
+provider "time" {}
+resource "time_rotating" "ninety_days" {
+    rotation_days = 90
+}
+
 resource "prefect_service_account" "sa_example" {
-    name = "beautiful-service-account-example"
+    name = "my-service-account"
+    api_key_expiration = time_rotating.ninety_days.rotation_rfc3339
```
parkedwards, line 8: this is the way we'll recommend practitioners handle key rotations, since the `time_rotating` resource handles properly exporting an rfc3339 compliant datetime string

jamiezieziula (approved):

internal/provider/resources/service_account.go
```diff
@@ -93,18 +104,13 @@ func (r *ServiceAccountResource) Schema(_ context.Context, _ resource.SchemaRequ
...
 			},
 			"account_id": schema.StringAttribute{
+				Optional:    true,
 				Computed:    true,
 				CustomType:  customtypes.UUIDType{},
 				Description: "Account UUID, defaults to the account set in the provider",
 			},
-			// @TODO: Do we actually need 'account_role_id' defined as an attribute here?
-			// or just only one of account_role_id or account_role_name?
-			"account_role_id": schema.StringAttribute{
```
parkedwards, line 102: account_role_id will be a request parameter upon Create, but since it doesn't get returned from the API (nor does the id have have meaning for the practitioner here), we'll omit it from the resource schema

in a followup, we'll use an optional `account_role_name` attribute to do an `account_role` lookup to find the id + pass it into the Create call, which will remain under the hood / out of view from the practitioner

internal/provider/resources/service_account.go
```diff
@@ -244,10 +242,7 @@ func (r *ServiceAccountResource) Read(ctx context.Context, req resource.ReadRequ
 		return
 	}
 
-	resp.Diagnostics.Append(copyServiceAccountToModel(ctx, serviceAccount, &model)...)
```
parkedwards, line 247: we dont return a diagnostic in the helper, so we dont need to pass it into `Append`

PR #75, opened 2023-10-27
feat: `prefect_workspace_access` Resource

### LINKED ISSUE #74: `prefect_workspace_access` resource
will unify workspace access operations for users, service accounts, and eventually teams

no datasource or resource import needed, as the primary write endpoint is an Upsert -- so, if a practitioner pulls in an existing service account and wants to use its workspace access, they can just define it in TF and it will overwrite any existing access links

parkedwards:
resolves https://github.com/PrefectHQ/terraform-provider-prefect/issues/74

This PR adds a new resource called `prefect_workspace_access`, which links an `accessor` (can be a user, service account, or any future type) with a workspace role

this resource / client are slightly different from the rest in that it's polymorphic in nature, so there are some shared data models (eg. `api.WorkspaceAccess`), which will conditionally present a non-nil value for `userID` or `botID` depending on the `accessorType`.  

Additionally, we've opted to skip supporting a datasource or resource state import in this instance; the workspace access API takes upserting as the primary write path via the combination of an accessor <> workspace, so any future practitioner can just redefine an existing access relationship in TF without worrying about having to import anything

The `accessorType` is restricted on the resource schema by a `OneOf()` string validator - therefore, this PR also introduces a new go dependency called [terraform-plugin-framework-validators](github.com/hashicorp/terraform-plugin-framework-validators/stringvalidator)

Usage example (included in the PR)
```hcl
data "prefect_workspace_role" "developer" {
	name = "Developer"
}
data "prefect_workspace" "prd" {
	id = "<workspace uuid>"
}
resource "prefect_service_account" "bot" {
	name = "a-cool-bot"
}
resource "prefect_workspace_access" "bot_access" {
	accessor_type = "SERVICE ACCOUNT"
	accessor_id = prefect_service_account.bot.id
	workspace_id = data.prefect_workspace.prd.id
	workspace_role_id = data.prefect_workspace_role.developer.id
}
```

Applying this to our acceptance testing environment shows the link between the workspace and Service Account in the `Workspace Sharing` section

image

### DIFF
--- a/examples/workspace_access.tf
+++ b/examples/workspace_access.tf
@@ -8,7 +8,7 @@ resource "prefect_service_account" "bot" {
 	name = "a-cool-bot"
 }
 resource "prefect_workspace_access" "bot_access" {
-	accessor_type = "SERVICE ACCOUNT"
+	accessor_type = "SERVICE_ACCOUNT"
 	accessor_id = prefect_service_account.bot.id
 	workspace_id = data.prefect_workspace.prd.id
 	workspace_role_id = data.prefect_workspace_role.developer.id
--- a/examples/workspace_access.tf
+++ b/examples/workspace_access.tf
@@ -1,0 +1,15 @@
+data "prefect_workspace_role" "developer" {
+	name = "Developer"
+}
+data "prefect_workspace" "prd" {
+	id = "<workspace uuid>"
+}
+resource "prefect_service_account" "bot" {
+	name = "a-cool-bot"
+}
+resource "prefect_workspace_access" "bot_access" {
+	accessor_type = "SERVICE ACCOUNT"
+	accessor_id = prefect_service_account.bot.id
+	workspace_id = data.prefect_workspace.prd.id
+	workspace_role_id = data.prefect_workspace_role.developer.id
+}
--- a/go.mod
+++ b/go.mod
@@ -5,6 +5,7 @@ go 1.20
 require (
 	github.com/google/uuid v1.3.1
 	github.com/hashicorp/terraform-plugin-framework v1.4.2
+	github.com/hashicorp/terraform-plugin-framework-validators v0.12.0
 	github.com/hashicorp/terraform-plugin-go v0.19.0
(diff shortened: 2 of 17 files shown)


examples/workspace_access.tf
```diff
@@ -0,0 +1,15 @@
...
+}
+resource "prefect_service_account" "bot" {
+	name = "a-cool-bot"
+}
+resource "prefect_workspace_access" "bot_access" {
+	accessor_type = "SERVICE ACCOUNT"
+	accessor_id = prefect_service_account.bot.id
+	workspace_id = data.prefect_workspace.prd.id
+	workspace_role_id = data.prefect_workspace_role.developer.id
+}
```
jamiezieziula, line 15: can you introduce terraform fmt as a precommit dep or as a PR linter?

examples/workspace_access.tf
```diff
@@ -0,0 +1,15 @@
...
+	name = "Developer"
+}
+data "prefect_workspace" "prd" {
+	id = "<workspace uuid>"
+}
+resource "prefect_service_account" "bot" {
+	name = "a-cool-bot"
+}
+resource "prefect_workspace_access" "bot_access" {
+	accessor_type = "SERVICE ACCOUNT"
```
jamiezieziula, line 11: should this be `SERVICE_ACCOUNT` as seen below in 
```
const ServiceAccount string = "SERVICE_ACCOUNT"
```

jamiezieziula (approved):

examples/workspace_access.tf
jamiezieziula: ignore i see that you did in https://github.com/PrefectHQ/terraform-provider-prefect/pull/77/files

parkedwards: great minds. i was like...something feels like it's missing

examples/workspace_access.tf
parkedwards: indeed it should, good catch

PR #111, opened 2023-11-07
fix: update variable resource name to be valid

jamiezieziula:


docs/resources/variable.md
```diff
@@ -44,8 +44,8 @@ resource "prefect_variable" "example" {
 Import is supported using the following syntax:
 
 ```shell
-# Prefect Variables can be imported via name in the form `name/name-of-variable`
-terraform import prefect_variable.example name/name-of-variable
+# Prefect Variables can be imported via name in the form `name/name_of_variable`
+terraform import prefect_variable.example name/name_of_variable
```
parkedwards, line 48: does it not like hyphens?

jamiezieziula: correct - lower case and underscores only

parkedwards (approved):

docs/resources/variable.md
parkedwards: ah wild, TIL

PR #115, opened 2023-11-07
fix: set default empty list for `prefect_variable.tags`

### LINKED ISSUE #110: Error when creating a `prefect_variable`, issue with `tags`
`version = "0.0.0-alpha5"`

When attempting to create a variable with the following config, i receive the below error:
```
resource "prefect_variable" "example" {
  name  = "my_variable"
  value = "variable value goes here"
}
```

```
│ Error: Provider produced inconsistent result after apply
│ 
│ When applying changes to prefect_variable.example, provider "provider[\"registry.terraform.io/prefecthq/prefect\"]" produced an unexpected new value:
│ .tags: was null, but now cty.ListValEmpty(cty.String).
│ 
│ This is a bug in the provider, which should be reported in the provider's own issue tracker.
```

The variable is created but when I run a subsequent plan/apply, I see this proposed change which results in the same error as listed above
```
  # prefect_variable.example is tainted, so must be replaced
-/+ resource "prefect_variable" "example" {
      ~ created = "2023-11-07T19:42:38Z" -> (known after apply)
      ~ id      = "afca6fee-8ff3-49e4-80bb-fd363902095f" -> (known after apply)
        name    = "my_variable"
      - tags    = [] -> null
      ~ updated = "2023-11-07T19:42:38Z" -> (known after apply)
        # (1 unchanged attribute hidden)
```

parkedwards:
resolves https://github.com/PrefectHQ/terraform-provider-prefect/issues/110

with this configuration:
```terraform
resource "prefect_variable" "example" {
  name  = "my_variable"
  value = "variable value goes here"
  workspace_id = data.prefect_workspace.evergreen.id
}
```

this is the plan that is generated (and applies successfully)

image

adding new tags, modifying tags, and resetting the resource to be tag-less all apply correctly

### DIFF
--- a/internal/provider/resources/variable.go
+++ b/internal/provider/resources/variable.go
@@ -6,13 +6,16 @@ import (
 	"strings"
 
 	"github.com/google/uuid"
+	"github.com/hashicorp/terraform-plugin-framework/attr"
 	"github.com/hashicorp/terraform-plugin-framework/diag"
 	"github.com/hashicorp/terraform-plugin-framework/path"
 	"github.com/hashicorp/terraform-plugin-framework/resource"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/listdefault"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema/planmodifier"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema/stringplanmodifier"
 	"github.com/hashicorp/terraform-plugin-framework/types"
+	"github.com/hashicorp/terraform-plugin-framework/types/basetypes"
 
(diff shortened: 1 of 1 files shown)


internal/provider/resources/variable.go
```diff
@@ -120,6 +125,8 @@ func (r *VariableResource) Schema(_ context.Context, _ resource.SchemaRequest, r
 				Description: "Tags associated with the variable",
 				ElementType: types.StringType,
 				Optional:    true,
+				Computed:    true,
+				Default:     listdefault.StaticValue(defaultEmptyTagList),
```
parkedwards, line 129: since this is a list type, there are these special constructs you need to use to set default values here

https://developer.hashicorp.com/terraform/plugin/framework/handling-data/attributes/list#plan-modification

jamiezieziula (approved):

PR #152, opened 2024-03-27
feat: Add WorkQueueClient to the provider

jamiezieziula:

PR #153, opened 2024-03-27
feat: Creates a data source for a `work_queue`

jamiezieziula:


internal/provider/datasources/work_queue.go
```diff
@@ -0,0 +1,172 @@
...
+	// Populate the model from data source configuration and emit diagnostics on error
+	resp.Diagnostics.Append(req.Config.Get(ctx, &model)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	client, err := d.client.WorkQueues(model.AccountID.ValueUUID(), model.WorkspaceID.ValueUUID(), model.WorkPoolName.ValueString())
+	if err != nil {
+		resp.Diagnostics.AddError(
+			"Error creating variable client",
```
parkedwards, line 142:
```suggestion
			"Error creating work queue client",
```

internal/provider/datasources/work_queue.go
```diff
@@ -0,0 +1,172 @@
...
+			fmt.Sprintf("Could not create variable client, unexpected error: %s. This is a bug in the provider, please report this to the maintainers.", err.Error()),
```
parkedwards, line 143:
```suggestion
			"Error creating work queue client",
			fmt.Sprintf("Could not create work queue client, unexpected error: %s. This is a bug in the provider, please report this to the maintainers.", err.Error()),
```

internal/provider/datasources/work_queue_test.go
```diff
@@ -0,0 +1,50 @@
...
+	return `
+data "prefect_workspace" "evergreen" {
+	handle = "github-ci-tests"
+}
+data "prefect_work_pool" "evergreen" {
+	name = "evergreen-pool"
+	workspace_id = data.prefect_workspace.evergreen.id
+}
+data "prefect_work_queue" "evergreen" {
+	name = "evergreen-queue"
```
parkedwards, line 20: assuming this exists?

jamiezieziula: yes!

internal/provider/datasources/work_queue.go
jamiezieziula: ok I was curious about this - we reference `variable client` in a few other go files. I will update them on this branch

internal/provider/datasources/work_queue.go
```diff
@@ -0,0 +1,172 @@
...
+	"work_pool_name": schema.StringAttribute{
+		Computed:    true,
+		CustomType:  customtypes.TimestampType{},
+		Description: "The name of the work pool that the work queue belongs to",
+	},
+	"name": schema.StringAttribute{
+		Computed:    true,
+		Description: "Name of the work queue",
+		Optional:    true,
+	},
```
parkedwards, line 95: so when i look at the fetching logic:

```go
client, err := d.client.WorkQueues(model.AccountID.ValueUUID(), model.WorkspaceID.ValueUUID(), model.WorkPoolName.ValueString())
```

^ when we initialize this work queues client, we're looking for three attributes on the datasource:

```hcl
data "prefect_work_queue" "foobar" {
#  ...
  work_pool_name = "my-pool"
  account_id = "ffff-ffff-ffff"
  workspace_id = "ffff-ffff-ffff"
}
```

so first thing is that we need to add `account_id` and `workspace_id` to the attributes map (you have it in the struct already as `AccountID` and `WorkspaceID`)

however, `account_id` and `workspace_id` can be OPTIONAL here, since the `client.WorkQueues()` factory function will fall back to the account_id / workspace_id set on the provider (this is a pattern we've adopted early on - allows the TF operator to create a datasource with an override to a different account/workspace if needed).  Take a look at `datasources/variable.go` as an example

but you will need the `work_pool_name` to be required, i would imagine

parkedwards: second, on the actual `Get()` call:

```
queue, err := client.Get(ctx, model.Name.ValueString())
```

^ we need the `name` key to be present, always.  So I think we will want to make this required, instead of optional

for example, on the Get() call, you pass it into the URL path - so we should expect this to always exist when someone creates a work_queue datasource bc it is the main identifier:
https://github.com/PrefectHQ/terraform-provider-prefect/pull/152/files#diff-e146e91f55659265a162378cc118455b6f0a7027ab74c39a8860ebd153dcf5a6R115-R116

parkedwards: so if im understanding this correctly, we should always expect, at minimum, these keys to exist on a work_queue datasource - and if so, we need to make them required

```hcl
data "prefect_work_queue" "foobar" {
#  ...
  work_pool_name = "my-pool"
  name = "my-queue"
}
```

PR #158, opened 2024-04-16
minor: fix docs around authenticating provider with environment variables

youcandanch:
Summary
-------

Spots in the documentation are inconsistent about whether to use PREFECT_API_KEY or PREFECT_CLOUD_API_KEY when passing the provider authentication config via environment variables. This PR consolidates documentation around this to use PREFECT_API_KEY, which is what the provider is currently looking for [here](https://github.com/PrefectHQ/terraform-provider-prefect/blob/a5ca0d39fd2ec3b378525d94f217114fc38580b7/internal/provider/provider.go#L146).

### DIFF
--- a/_about/CONTRIBUTING.md
+++ b/_about/CONTRIBUTING.md
@@ -98,7 +98,7 @@ will run [TF acceptance tests](https://developer.hashicorp.com/terraform/plugin/
 
 ```shell
 export PREFECT_API_URL=https://api.prefect.cloud
-export PREFECT_CLOUD_API_KEY=<secret>
+export PREFECT_API_KEY=<secret>
 export PREFECT_CLOUD_ACCOUNT_ID=<uuid>
 
 make testacc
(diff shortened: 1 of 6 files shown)


parkedwards:
ah thanks for the PR @youcandanch -- you are correct, it should be the other way around. the correct env var is `PREFECT_API_KEY`, and it looks like we've mistakenly documented this var as `PREFECT_CLOUD_API_KEY`

if you could update all doc references to `PREFECT_API_KEY`, that would be amazing

youcandanch:
@parkedwards yeaaaaah kinda figured once I changed it and the acceptance tests immediately gave me the virtual bird. Give me a sec and I'll update!

youcandanch:
Acceptance tests dependent on https://github.com/PrefectHQ/terraform-provider-prefect/pull/158, another casualty of the underscore war.

parkedwards (approved):
ill fix the labeler and docs workflow jobs, they're not being permitted bc it's coming from an external fork. thanks for submitting!

Issue #171, opened 2024-05-23
Blocks Support: Add BlockType HTTP client + struct model


parkedwards:
https://www.notion.so/prefect/Design-Block-support-in-the-Terraform-Provider-ae2ea25a18864de9b51b8aa06af9486b?pvs=4

In order to create a Block, we need a `block_type.id` and a `block_schema.id`.  So we'll want an HTTP client for `block_type`, so we can fetch one

Things we'll probably need:
* `api.BlockType` (struct) - in internal/api
* `BlockTypeClient interface{}`- in internal/client
* `func (c *BlockTypeClient) GetBySlug(ctx context.Context, slug string) (*api.BlockType, error)` - in internal/client

https://app.prefect.cloud/api/docs#tag/Block-types/operation/read_block_type_by_slug_api_accounts__account_id__workspaces__workspace_id__block_types_slug__slug__get

PR #176, opened 2024-05-23
Add BlockType, BlockTypeClient

mitchnielsen:
- Adds BlockType struct in internal/api
- Adds BlockTypeClient interface in internal/client
- Adds GetBySlug in internal/client

Closes
https://github.com/PrefectHQ/terraform-provider-prefect/issues/171

### DIFF
--- a/internal/client/blocks.go
+++ b/internal/client/blocks.go
@@ -1,0 +1,68 @@
+package client
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+
+	"github.com/google/uuid"
+	"github.com/prefecthq/terraform-provider-prefect/internal/api"
+)
+
+var _ = api.BlockTypeClient(&BlockTypeClient{})
+
+// BlockTypeClient is a client for working with block types.
+type BlockTypeClient struct {
+	hc          *http.Client
+	routePrefix string
+	apiKey      string
+}
+
+// BlockTypes returns a BlockTypeClient.
+//
+//nolint:ireturn // required to support PrefectClient mocking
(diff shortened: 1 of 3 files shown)


mitchnielsen:
Putting in `draft` state since this is still under development.

internal/client/blocks.go
```diff
@@ -0,0 +1,68 @@
...
+	if workspaceID == uuid.Nil {
+		workspaceID = c.defaultWorkspaceID
+	}
+
+	return &BlockTypeClient{
+		hc:          c.hc,
+		apiKey:      c.apiKey,
+		routePrefix: getWorkspaceScopedURL(c.endpoint, accountID, workspaceID, "block_types"),
+	}, nil
+}
```
parkedwards, line 39: one quick thing - for every new client factory, we'll want to add add the factory's function signature into the `PrefectClient` interface here:

https://github.com/PrefectHQ/terraform-provider-prefect/blob/main/internal/api/client.go#L8-L20

that way, we can have our client types for when we instantiate the actual client -> pass into our provider at boot time

https://github.com/PrefectHQ/terraform-provider-prefect/blob/main/internal/provider/provider.go#L194-L198

mitchnielsen: Makes sense! Updated in 88e3bb6.

parkedwards (approved):
fire. out of curiosity, what did you have to change to fix that `Client` interface error?

mitchnielsen:
> fire. out of curiosity, what did you have to change to fix that `Client` interface error?

Just a typo, I had written `BlockType` instead of `BlockTypes`

PR #178, opened 2024-05-30
feat(blocks): add BlockDocument HTTP client

### LINKED ISSUE #173: Blocks Support: Add BlockDocument HTTP client + struct model

A `block_document` is the **actual** API object that holds the user's value

Things we'll probably need:

- api.BlockDocument (struct) - in internal/api (the API response object0
- api.BlockDocumentCreateRequest
- api.BlockDocumentUpdateRequest
- BlockDocumentClient interface{}- in internal/client
- func (c *BlockSchemaClient) Get(ctx context.Context, ID uuid.UUID) (*api.BlockDocument, error) - in internal/client
- func (c *BlockSchemaClient) Create(ctx context.Context, payload *api.BlockDocumentCreateRequest) (*api.BlockDocument, error) - in internal/client
- func (c *BlockSchemaClient) Update(ctx context.Context, ID uuid.UUID, payload *apiBlockDocumentUpdateRequest) (*api.BlockDocument, error) - in internal/client
- func (c *BlockSchemaClient) Delete(ctx context.Context, ID uuid.UUID) (error) - in internal/client

note that we can also include a `?include_secrets=true` query param, so that we can always hydrate the TF state with the actual value

parkedwards:
resolves https://github.com/PrefectHQ/terraform-provider-prefect/issues/173

adds:
* `BlockDocumentClient` - with Get, Create, Update, and Delete methods
* a subset of ^ client for `/access`, to manage attached ACL objects - specifically:
  * GetACL (GET /block_documents/{id}/access)
  * UpdateACL (PUT /block_documents/{id}/access)

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -18,7 +18,6 @@ linters:
     - errchkjson
     - errname
     - errorlint
-    - execinquery
     - exhaustive
     - exportloopref
     - forbidigo
@@ -32,7 +31,7 @@ linters:
     - godox
     - goheader
     - goimports
-    - gomnd
+    - mnd
     - gomoddirectives
     - gomodguard
     - goprintffuncname
@@ -76,7 +75,6 @@ linters:
     - unconvert
     - unused
     - usestdlibvars
-    - varnamelen
     - wastedassign
     - whitespace
     - wrapcheck
--- a/internal/api/block_documents.go
+++ b/internal/api/block_documents.go
@@ -2,9 +2,6 @@ package api
 
 import (
 	"context"
-	"encoding/json"
-	"errors"
-	"fmt"
 
 	"github.com/google/uuid"
 )
@@ -50,60 +47,6 @@ type BlockDocumentUpdate struct {
 	MergeExistingData bool       `json:"merge_existing_data"`
 }
 
-// AccessControlList is a custom type that represents
-// an API response where the value can be:
-// []uuid.UUID - a list of IDs
-// ["*"] - a wildcard, meaning "all".
-//
-// nolint:musttag // we have custom marshal/unmarshal logic for this type
-type AccessControlList struct {
-	IDs []uuid.UUID
-	All bool
-}
-
-// Custom JSON marshaling for AccessControlList
-// so we can return []uuid.UUID or ["*"] back to the API.
-func (acl AccessControlList) MarshalJSON() ([]byte, error) {
-	if acl.All {
-		data, err := json.Marshal([]string{"*"})
-		if err != nil {
-			return nil, fmt.Errorf("failed to marshal wildcard ACL: %w", err)
-		}
-
-		return data, nil
-	}
-
-	data, err := json.Marshal(acl.IDs)
-	if err != nil {
-		return nil, fmt.Errorf("failed to marshal ID slice ACL: %w", err)
-	}
-
-	return data, nil
-}
-
-// Custom JSON unmarshaling for AccessControlList
-// so we can accept []uuid.UUID or ["*"] from the API
-// in a structured format.
-func (acl *AccessControlList) UnmarshalJSON(data []byte) error {
-	var ids []uuid.UUID
-	if err := json.Unmarshal(data, &ids); err == nil {
-		acl.IDs = ids
-		acl.All = false
-
-		return nil
-	}
-
-	var all []string
-	if err := json.Unmarshal(data, &all); err == nil && len(all) == 1 && all[0] == "*" {
-		acl.All = true
-		acl.IDs = nil
-
-		return nil
-	}
-
-	return errors.New("invalid AccessControlList format")
-}
-
 // BlockDocumentAccessReplace is the "update" request payload
 // to modify a block document's current access control levels,
 // meaning it contains the list of actors/teams + their respective access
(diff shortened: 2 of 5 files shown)


.golangci.yaml
```diff
@@ -18,7 +18,6 @@ linters:
     - errchkjson
     - errname
     - errorlint
-    - execinquery
```
parkedwards, line 21: deprecated

.golangci.yaml
```diff
@@ -76,7 +75,6 @@ linters:
     - unconvert
     - unused
     - usestdlibvars
-    - varnamelen
```
parkedwards, line 79: [this is an annoying rule](https://github.com/blizzy78/varnamelen); i can understand the original motivation, but in practice, i'd like to use short variable names like `id`

.golangci.yaml
```diff
@@ -32,7 +31,7 @@ linters:
     - godox
     - goheader
     - goimports
-    - gomnd
```
parkedwards, line 35: the newer version of this is called `mnd`

internal/api/block_documents.go
```diff
@@ -0,0 +1,142 @@
...
+	UpdateACL(ctx context.Context, id uuid.UUID, payload BlockDocumentAccessReplace) error
+}
+
+type BlockDocument struct {
+	BaseModel
+	Name *string `json:"name"` // names are optional for anonymous blocks
+	Data string  `json:"data"`
+
+	BlockSchemaID uuid.UUID `json:"block_schema_id"`
+	// BlockSchema   *BlockSchema `json:"block_schema"`
```
parkedwards, line 28: forthcoming, when the BlockSchema API and client are added

internal/api/block_documents.go
```diff
@@ -0,0 +1,142 @@
...
+// AccessControlList is a custom type that represents
+// an API response where the value can be:
+// []uuid.UUID - a list of IDs
+// ["*"] - a wildcard, meaning "all".
+//
+// nolint:musttag // we have custom marshal/unmarshal logic for this type
+type AccessControlList struct {
+	IDs []uuid.UUID
+	All bool
+}
```
parkedwards, line 62: basically, i've set up this struct type to allow us to serialize and better handle this varied type from the API

internal/api/block_documents.go
```diff
@@ -0,0 +1,142 @@
...
+// BlockDocumentAccessReplace is the "update" request payload
+// to modify a block document's current access control levels,
+// meaning it contains the list of actors/teams + their respective access
+// to a given block document.
+type BlockDocumentAccessReplace struct {
+	ManageActorIDs AccessControlList `json:"manage_actor_ids"`
+	ViewActorIDs   AccessControlList `json:"view_actor_ids"`
+	ManageTeamIDs  []uuid.UUID       `json:"manage_team_ids"`
+	ViewTeamIDs    []uuid.UUID       `json:"view_team_ids"`
+}
```
parkedwards, line 116: might be worth noting, but our access API only takes a PUT call - meaning every call to this endpoint will declaratively set the latest access list for the block document in question

internal/client/block_documents.go
```diff
@@ -0,0 +1,212 @@
...
+
+var _ = api.BlockDocumentClient(&BlockDocumentClient{})
+
+type BlockDocumentClient struct {
+	hc          *http.Client
+	apiKey      string
+	routePrefix string
+}
+
+// BlockDcouments is a factory that initializes and returns a BlockDocumentClient.
```
mitchnielsen, line 23:
```suggestion
// BlockDocuments is a factory that initializes and returns a BlockDocumentClient.
```

internal/api/block_documents.go
mitchnielsen: Looks good, two quick questions:

- The API seems to refer to these as `access_control` objects. Would `type AccessControl struct` make more sense here? Maybe not, but it threw me off a bit when searching the API docs.
- Looks like `access_control` is part of the payload for [Set Deployment Access](https://app.prefect.cloud/api/docs#tag/Deployments/operation/set_deployment_access_api_accounts__account_id__workspaces__workspace_id__deployments__id__access_put) as well. Would it make sense to put this struct elsewhere, in a common/helpers file? (happy to merge as-is now and we can move it when we work on support for Deployments).

mitchnielsen (changes requested):
Awesome work Eddie, looks great. Two quick questions and one tiny suggestion.

internal/api/block_documents.go
parkedwards: > The API seems to refer to these as access_control objects. Would type AccessControl struct make more sense here? Maybe not, but it threw me off a bit when searching the API docs.

I'm down for that - although when I was skimming our API spec, i interpreted "access_control" = the entire payload, including the top-level keys for `manage`, `view`, etc., which each point to a list of IDs / `"*"`.  And then this here custom type would be for the weird, varied type holding the actual list of IDs (or the optional `"*"` string).  

> Looks like access_control is part of the payload for Set Deployment Access as well. Would it make sense to put this struct elsewhere, in a common/helpers file? (happy to merge as-is now and we can move it when we work on support for Deployments).

For sure. let me actually just create a helper module in `api` package and move this type there

mitchnielsen: > I'm down for that - although when I was skimming our API spec, i interpreted "access_control" = the entire payload, including the top-level keys for manage, view, etc., which each point to a list of IDs / "*". And then this here custom type would be for the weird, varied type holding the actual list of IDs (or the optional "*" string).

Makes sense to me as well, happy to leave it as it is 👍🏼 

> For sure. let me actually just create a helper module in api package and move this type there

Sounds great.

internal/api/helpers.go
```diff
@@ -0,0 +1,78 @@
...
+// AccessActorID is a custom type that represents
+// an API response where the value can be:
+// uuid.UUID - a single ID
+// "*" - a wildcard string, meaning "all".
+//
+// nolint:musttag // we have custom marshal/unmarshal logic for this type
+type AccessActorID struct {
+	ID  *uuid.UUID
+	All bool
+}
```
parkedwards, line 19: tiny correction - this custom type should reflect either:
1. a _**single**_ uuid
2. a `"*"` wildcard string

on PUT calls, we'll pass a `[]AccessActorID` (or a list of ^)

```json
{
  "manage_actor_ids": [
    "*"
  ],
  "view_actor_ids": [
    "*"
  ],
  "manage_team_ids": [],
  "view_team_ids": []
}
```

on GET calls, we'll hydrate a list of `[]ObjectActorAccess`, each of which will have an `ID: AccessActorID` <- this is a single UUID/wildcard, since the encapsulating data model is what gets put into a list

```json
{
  "manage_actors": [
    {
      "id": "*",
      "name": "*",
      "email": "*",
      "type": "*"
    }
  ],
  "view_actors": [
    {
      "id": "*",
      "name": "*",
      "email": "*",
      "type": "*"
    }
  ]
}
```

mitchnielsen:
🎉 🚢

mitchnielsen (approved):

PR #180, opened 2024-05-31
Add BlockSchema HTTP client and struct model

### LINKED ISSUE #172: Blocks Support: Add BlockSchema HTTP client + struct model

In order to create a Block, we need a block_type.id and a block_schema.id. So we'll want an HTTP client for block_schema, so we can fetch one

Things we'll probably need:

* `api.BlockSchema` (struct) - in internal/api
* `api.BlockSchemaFilter` (struct) - in internal/api -- this is the filter payload on the search request
* `BlockSchemaClient interface{}`- in internal/client
* `func (c *BlockSchemaClient) List(ctx context.Context, blockTypeSlugs []string) ([]*api.BlockSchema, error)` - in internal/client

https://app.prefect.cloud/api/docs#tag/Block-schemas/operation/read_block_schemas_api_accounts__account_id__workspaces__workspace_id__block_schemas_filter_post

mitchnielsen:
Closes https://github.com/PrefectHQ/terraform-provider-prefect/issues/172


mitchnielsen:
Was getting an error from `pre-commit`:

```
$ g cm
Terraform fmt........................................(no files to check)Skipped
go fmt...................................................................Passed
go vet...................................................................Passed
go imports...............................................................Passed
validate toml........................................(no files to check)Skipped
golangci-lint............................................................Failed
- hook id: golangci-lint
- exit code: 1

internal/client/block_documents.go:178:46: the given struct should be annotated with the `json` tag (musttag)
        if err := json.NewDecoder(resp.Body).Decode(&blockDocumentAccess); err != nil {
                                                    ^
```

I suspect this came from https://github.com/PrefectHQ/terraform-provider-prefect/pull/178, but surprised pre-commit / CI didn't complain about it.

Do you get this error locally, too @parkedwards? I did kind of wonder how the unmarshaling would work since we don't have struct tags on those types 🤔

parkedwards:
@mitchnielsen what the heck - so when I was dev-ing on #178 , I _**did**_ run into this linting error

[the issue is actually with this struct](https://github.com/PrefectHQ/terraform-provider-prefect/pull/178/files#diff-3e8b73f5901bb9b41436eb294ebc11f27d7f3f93d07c779b0be7a771af8a3bdfR10-R19), but it's (1) nested in the `BlockDocumentAccess`, (2) I added a nolint comment, and (3) the `BlockDocumentAccess` struct does have struct tags

i think i upgraded my golangci-lint binary to get around this - what version are you running?

```
➜ golangci-lint --version
golangci-lint has version 1.59.0 built with go1.22.3 from 2059b18 on 2024-05-25T11:38:08Z
```

mitchnielsen:
> i think i upgraded my golangci-lint binary to get around this - what version are you running?

Ah interesting, I have this version from Homebrew (which it says is the latest available to it):

`golangci-lint has version 1.58.1 built with go1.22.3 from dc28153 on 2024-05-07T23:17:27Z`

I've been thinking we might benefit from using [`asdf`](https://asdf-vm.com/) or [`mise`](https://mise.jdx.dev/) so we can track the tool versions in a file that we commit to the repo so we're all on the same page.

mitchnielsen:
https://golangci-lint.run/product/changelog/#1582:

> musttag: from 0.12.1 to 0.12.2

https://github.com/go-simpler/musttag/releases/tag/v0.12.2:

https://github.com/go-simpler/musttag/commit/5b5f6e32a8fa90b185daf7f30aeed01157a26656 fix: check whether a pointer-type implements a Marshaler interface (https://github.com/go-simpler/musttag/pull/94)

That seems to be what fixed it, since updating to `golangci-lint` 1.59.0 fixed that `musttag` check 👍🏼

parkedwards:
ok phew, my sanity is restored 💆

internal/api/block_documents.go
```diff
@@ -21,8 +21,8 @@ type BlockDocument struct {
 	Name *string `json:"name"` // names are optional for anonymous blocks
 	Data string  `json:"data"`
 
-	BlockSchemaID uuid.UUID `json:"block_schema_id"`
-	// BlockSchema   *BlockSchema `json:"block_schema"`
+	BlockSchemaID uuid.UUID    `json:"block_schema_id"`
+	BlockSchema   *BlockSchema `json:"block_schema"`
```
parkedwards, line 25: TY!

internal/api/block_schemas.go
```diff
@@ -0,0 +1,48 @@
...
+
+// BlockTypeID represents a block type ID.
+type BlockTypeID struct {
+	Any []uuid.UUIDs `json:"any_"`
+}
+
+// BlockSchemas represents a block schema.
+type BlockSchemas struct {
+	BlockTypeID
+}
```
parkedwards, line 34: ooc, how are these guys getting used?

internal/client/block_schemas.go
```diff
@@ -0,0 +1,75 @@
...
+
+	return &BlockSchemaClient{
+		hc:          c.hc,
+		apiKey:      c.apiKey,
+		routePrefix: getWorkspaceScopedURL(c.endpoint, accountID, workspaceID, "block_schemas"),
+	}, nil
+}
+
+// List gets a list of BlockSchemas for a given list of block type slugs.
+func (c *BlockSchemaClient) List(ctx context.Context, blockTypeSlugs []uuid.UUIDs) ([]*api.BlockSchema, error) {
```
parkedwards, line 41:
```suggestion
func (c *BlockSchemaClient) List(ctx context.Context, blockTypeIDs []uuid.UUIDs) ([]*api.BlockSchema, error) {
```

internal/api/block_schemas.go
mitchnielsen: Ah, meant to delete this since I used the nested struct instead. The nested struct was the only way I've seen so far to still provide struct tags (in this example, I'm not sure how I'd add a struct tag to `BlockTypeID`).

Fixed in 93fe963.

internal/api/block_schemas.go
```diff
@@ -0,0 +1,38 @@
...
+
+// BlockSchemaClient is a client for working with block schemas.
+type BlockSchemaClient interface {
+	List(ctx context.Context, blockTypeIDs []uuid.UUIDs) ([]*BlockSchema, error)
+}
+
+// BlockSchema is a representation of a block schema.
+type BlockSchema struct {
+	BaseModel
+	BlockType
```
parkedwards, line 17: im a little hazy with embeddings <> json serialization, but if the API payload looks something like:

```json
{
  "id": "06c34bb7-ac74-4dbb-bcdb-407daa305e74",
  "created": "2024-04-05T19:55:34.265851+00:00",
  "updated": "2024-04-05T19:55:34.265867+00:00",
  "checksum": "sha256:273dfe712133cba3cdb092d88741da37e33cee892bfcf4fcf2895bdcee49a965",
  "fields": {...},
  "block_type_id": "063573f9-e491-48a4-8b2e-c1905a890c7e",
  "block_type": {
    "id": "063573f9-e491-48a4-8b2e-c1905a890c7e",
    "created": "2023-11-09T20:32:13.680977+00:00",
    "updated": "2023-11-09T20:32:13.681003+00:00",
    "name": "Slack Webhook",
    "slug": "slack-webhook",
    "logo_url": "https://cdn.sanity.io/images/3ugk85nk/production/c1965ecbf8704ee1ea20d77786de9a41ce1087d1-500x500.png",
    "documentation_url": "https://docs.prefect.io/api-ref/prefect/blocks/notifications/#prefect.blocks.notifications.SlackWebhook",
    "description": "Enables sending notifications via a provided Slack webhook.",
    "code_example": "Load a saved Slack webhook and send a message:\n```python\nfrom prefect.blocks.notifications import SlackWebhook\n\nslack_webhook_block = SlackWebhook.load(\"BLOCK_NAME\")\nslack_webhook_block.notify(\"Hello from Prefect!\")\n```",
    "is_protected": false
  },
  "capabilities": [
    "notify"
  ],
  "version": "2.16.9"
}
```

should we be setting a `BlockType` property here instead of embedding?  

```golang
type BlockSchema struct {
	BaseModel
	BlockType    BlockType   `json:"block_type"`
        BlockTypeID uuid.UUID   `json:"block_type_id"`
...
}
```

mitchnielsen: Potentially, although I figured this would work because we're already using that approach with `BaseModel`:

```go
// internal/api/base_model.go

// BaseModel is embedded in all other types and defines fields
// common to all Prefect data models.
type BaseModel struct {
	ID      uuid.UUID  `json:"id"`
	Created *time.Time `json:"created"`
	Updated *time.Time `json:"updated"`
}
```

🤔

mitchnielsen: I see, I think you're right because embeddeding doesn't put a struct inside of it - it just inherits the values of that type, which isn't what we want. Fixing.

mitchnielsen: Updated in 65fbb37.

parkedwards: right on

parkedwards (approved):

PR #185, opened 2024-05-31
Pin tool versions with 'mise'

### LINKED ISSUE #184: Manage tool versions for local development
As we saw in https://github.com/PrefectHQ/terraform-provider-prefect/pull/180#issuecomment-2141118647, tool versions being different on each developer's machine can lead to problems. It's helpful to manage these tool versions so developers are on the same page.

There are a couple tools that can help with this:

- https://asdf-vm.com: this one has been around for a while and was fairly popular, but it's "shim" approach meant that calls to the tools were fairly slow
- https://mise.jdx.dev: this is an alternative to `asdf` that drops the slow "shim" approach, and offers some [other benefits](https://mise.jdx.dev/dev-tools/comparison-to-asdf.html)

Let's try to manage the tool versions for this project and see if we find it helpful in this case. If so, we could consider doing the same for some of our other projects (where it makes sense).

_Note to self for future me: `mise` supports [GitHub CI Actions](https://mise.jdx.dev/continuous-integration.html#github-actions), which would help us use the same tool versions in CI as we use locally._

mitchnielsen:
Manage tool versions with mise

    Manages the tool versions used for development by adding .mise.toml and
    pinning the tool versions there. This will help keep the tool versions
    aligned between developer machines (and potentially CI in the future).

    https://mise.jdx.dev

Closes https://github.com/PrefectHQ/terraform-provider-prefect/issues/184

### DIFF
--- a/.mise.toml
+++ b/.mise.toml
@@ -1,0 +1,9 @@
+[tools]
+golangci-lint = '1.59.0'
+goreleaser = '1.26.2'
+pre-commit = '3.7.1'
+terraform = '1.8.4'
+
+# Installing Golang seems to be broken at the moment.
+# For now, install with `brew` per ./_about/CONTRIBUTING.md.
+# go-sdk = '1.21.10'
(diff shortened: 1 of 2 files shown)


jamiezieziula (approved):
nice, this is awesome!

parkedwards (approved):

PR #193, opened 2024-06-04
Implement Read method for BlockSource

### LINKED ISSUE #194: Blocks Support: Create prefect_block resource - Read()

mitchnielsen:
Implements BlockSource.Read method.

Closes https://github.com/PrefectHQ/terraform-provider-prefect/issues/194

### DIFF
--- a/internal/provider/resources/block.go
+++ b/internal/provider/resources/block.go
@@ -4,7 +4,9 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/google/uuid"
 	"github.com/hashicorp/terraform-plugin-framework-jsontypes/jsontypes"
+	"github.com/hashicorp/terraform-plugin-framework/path"
 	"github.com/hashicorp/terraform-plugin-framework/resource"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema/planmodifier"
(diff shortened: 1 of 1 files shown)


parkedwards:
@mitchnielsen i updated the linked issue in your PR description

internal/provider/resources/block.go
```diff
@@ -127,9 +129,64 @@ func (r *BlockResource) Create(ctx context.Context, req resource.CreateRequest,
...
+
+		return
+	}
+
+	state.ID = types.StringValue(block.ID.String())
+	state.Created = customtypes.NewTimestampPointerValue(block.Created)
+	state.Updated = customtypes.NewTimestampPointerValue(block.Updated)
+	state.Name = types.StringValue(*block.Name)
+	state.Type = types.StringValue(block.BlockType.Slug)
+	state.Data = jsontypes.NewNormalizedValue(block.Data)
```
mitchnielsen, line 183: We end up using this in other methods as well, like `Update`, so in a future PR I'll move this to a reusable function.

internal/provider/resources/block.go
```diff
@@ -127,9 +129,64 @@ func (r *BlockResource) Create(ctx context.Context, req resource.CreateRequest,
 }
 
 // Read refreshes the Terraform state with the latest data.
-//
-//nolint:revive // TODO: remove this comment when method is implemented
 func (r *BlockResource) Read(ctx context.Context, req resource.ReadRequest, resp *resource.ReadResponse) {
+	var state BlockResourceModel
```
mitchnielsen, line 133: Sometimes we call this `state`, sometimes `model`. The Terraform tutorial calls it `state`, which felt a little more intuitive for me when reading the rest of the function. But I'm good either way.

internal/provider/resources/block.go
```diff
@@ -127,9 +129,64 @@ func (r *BlockResource) Create(ctx context.Context, req resource.CreateRequest,
...
+	var blockID uuid.UUID
+	blockID, err = uuid.Parse(state.ID.ValueString())
+	if err != nil {
+		resp.Diagnostics.AddAttributeError(
+			path.Root("id"),
+			"Error parsing Block ID",
+			fmt.Sprintf("Could not parse block ID to UUID, unexpected error: %s", err.Error()),
+		)
+
+		return
```
parkedwards, line 165: it's worth mentioning here before we go too far down the rabbit hole - just like you have it here, anything that the user interfaces with here will be the `block`, whereas the API resources that constitute it are the `block_document` / `block_type` / `block_schema` etc.

parkedwards (approved):
nice!

internal/provider/resources/block.go
mitchnielsen: Good callout 👍🏼

internal/provider/resources/block.go
parkedwards: yeah, i think we should eventually do away with the `model` naming.  IMO we should only have `state` or `plan` variables, based on which framework function (eg. `req.State.Get()` or `req.Plan.Get()`) is doing the unmarshalling

mitchnielsen: Agreed 👍🏼 Opened https://github.com/PrefectHQ/terraform-provider-prefect/issues/195 so we can come back to that at some point.

Issue #183, opened 2024-05-31
Blocks Support: Create `prefect_block` resource - Tests + Docs


parkedwards:
https://www.notion.so/prefect/Design-Block-support-in-the-Terraform-Provider-ae2ea25a18864de9b51b8aa06af9486b?pvs=4

```hcl
resource "prefect_block" "datadog_api_key" {
	name = "datadog-api-key"
	type = "secret"
	data = jsonencode({
		value = var.datadog_api_key
	})
}
```

Located in `internal/provider/resources/block.go`

Also, be sure to include acceptance tests and docs
* `/internal/provider/resources/block_test.go`
* `/examples/resources/prefect_block`

mitchnielsen:
I'll pick this up for now as I'd like to learn how this part works, but feel free to reassign if needed.

parkedwards:
ah @mitchnielsen i just saw you grabbed it, sorry man. in https://github.com/PrefectHQ/terraform-provider-prefect/pull/196 i added some foundations for a `block_test.go` file, so there should be some stuff for us to look at for followup PRs

i couldn't quite get this to work, i think we might be missing some logic to gracefully handle the different URL schemas between open source and cloud https://github.com/PrefectHQ/terraform-provider-prefect/issues/192

mitchnielsen:
No worries at all, sounds good! And good to know.

PR #196, opened 2024-06-04
feat(blocks): add create/delete/import resource operations

### LINKED ISSUE #174: Blocks Support: Create `prefect_block` resource - Create(), Delete(), and ImportState() methods

```hcl
resource "prefect_block" "datadog_api_key" {
	name = "datadog-api-key"
	type = "secret"
	data = jsonencode({
		value = var.datadog_api_key
	})
}
```

Located in `internal/provider/resources/block.go`

We'll need to add a `Create()` and `Delete()` method on the pointer receiver for the `BlockResource` struct

We should be able to:
1. create a new `prefect_block`
2. have a `prefect_block` be unchanged if no attributes change

We'll need to add a `ImportState()` method on the pointer receiver for the `BlockResource` struct

Afterwards We should be able to:
1. import an existing `prefect_block`

parkedwards:
resolves https://github.com/PrefectHQ/terraform-provider-prefect/issues/174
relates to https://github.com/PrefectHQ/terraform-provider-prefect/issues/183 <- added some initial tests and example docs

# Testing
```hcl
terraform {
  required_providers {
    prefect = {
      source = "prefecthq/prefect"
    }
  }
}

provider "prefect" {
  endpoint = "https://api.prefect.dev"
  account_id = <id>
  api_key    = <key>
}

resource "prefect_block" "secret" {
  name      = "foo"
  type_slug = "secret"

  data = jsonencode({
    "value" : "bar"
  })

  workspace_id = <id>
}
```

```
terraform apply --auto-approve

terraform plan

terraform destroy --auto-approve
```


internal/api/block_documents.go
```diff
@@ -29,16 +29,13 @@ type BlockDocument struct {
 	BlockType     BlockType `json:"block_type"`
 
 	BlockDocumentReferences string `json:"block_document_references"`
-
-	IsAnonymous bool `json:"is_anonymous"`
 }
 
 type BlockDocumentCreate struct {
-	Name          *string   `json:"name"` // names are optional for anonymous blocks
+	Name          string    `json:"name"`
```
parkedwards, line 35: per our convo and confirming w the backend team, anonymous blocks are a server-side implementation and don't exist functionally for outside users

internal/provider/resources/block.go
```diff
@@ -27,9 +29,9 @@ type BlockResourceModel struct {
 	AccountID   customtypes.UUIDValue      `tfsdk:"account_id"`
 	WorkspaceID customtypes.UUIDValue      `tfsdk:"workspace_id"`
 
-	Name types.String         `tfsdk:"name"`
-	Type types.String         `tfsdk:"type"`
-	Data jsontypes.Normalized `tfsdk:"data"`
+	Name     types.String         `tfsdk:"name"`
+	TypeSlug types.String         `tfsdk:"type_slug"`
```
parkedwards, line 33: im a little undecided on this, but i was getting confused by `type` (since that's an API object for us), so maybe we specify this as a `type_slug` (or `type-slug`) on the TF config

mitchnielsen, line 33: Good change, ran across this when working on `Update` and wasn't sure what `Type` referred to here until looking at the schema description.

parkedwards: oh guess they'd be underscores, based on this:

https://developer.hashicorp.com/terraform/plugin/best-practices/naming#attribute-names

internal/provider/resources/block.go
```diff
@@ -116,16 +120,111 @@ func (r *BlockResource) Schema(_ context.Context, _ resource.SchemaRequest, resp
...
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	createdBlockDocument, err := blockDocumentClient.Create(ctx, api.BlockDocumentCreate{
+		Name:          config.Name.ValueString(),
+		Data:          data,
+		BlockSchemaID: latestBlockSchema.ID,
+		BlockTypeID:   latestBlockSchema.BlockTypeID,
+	})
```
parkedwards, line 210: full disclaimer, i dont love that i did this and im open to suggestions on making this better.  essentially:

1. `data` is a polymorphic JSON from the API
2. we need to type it on the terraform side - T[F does have a concept of dynamic types](https://developer.hashicorp.com/terraform/plugin/framework/handling-data/dynamic-data#considerations), but i opted not to use it to start bc it seemed a little heavy on the caveats. also not sure if it fits our use case here
3. since `data` is defined as JSON in the `.tf` code, we need to pass that forward to the API at some point on creates/updates - but since we already serialize the entire payload, i sort of needed to unmarshal it first to a `map[string]interface` or else it would get double encoded = API will reject it

some things im not sure of yet, but it's worth trying in a followup:
1. can we store the `data` key as a `string` in the golang struct before handing it off to a JSON type in TF? one nice thing is that we use a TF type library that will do some diffs against the normalized JSON, so it will intelligently tell us if there are only whitespace changes, for example
2. any way to exclude marshalling a struct property and pass it forward as-is?

internal/provider/resources/block_test.go
```diff
@@ -0,0 +1,39 @@
+package resources_test
+
+// func fixtureAccBlockCreate(name string) string {
```
parkedwards, line 3: commenting this out for now, as they will fail if we dont register the resource to the provider

internal/provider/resources/block.go
mitchnielsen: Thanks for the context. Thinking this over, too, and I'll post some links I'm still digesting. In the meantime though I'm happy to move forward with this and we can tweak as needed once we get more experience (and testing) in place.

-----

https://discuss.hashicorp.com/t/how-to-handle-json-input-into-data-resource-with-multiple-types-with-the-terraform-sdk/22896/2:

> The typical approach today is, I think, what you already tried: define a string attribute, often with a name ending in _json to be explicit about it, and then decode that value as JSON inside the provider code.

This seems to confirm your `string` attribute thought if I'm reading it correctly.

-----

https://github.com/hashicorp/terraform-plugin-go?tab=readme-ov-file#getting-started:

> When handling requests, [tfprotov5.DynamicValue](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-go/tfprotov5#DynamicValue) types should always have their [Unmarshal](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-go/tfprotov5#DynamicValue.Unmarshal) methods called; their properties should not be inspected directly. The [tftypes.Value](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-go/tfprotov5/tftypes#Value) returned from Unmarshal can be inspected to check [whether it is known](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-go/tfprotov5/tftypes#Value.IsKnown) and subsequently converted to a plain Go type using its [As](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-go/tfprotov5/tftypes#Value.As) method. As will return an error if the Value is not known.

Not sure how much crossover this has with the Terraform Dynamic Type docs you linked earlier, but this was mentioned in the first link I commented.

-----

Small aside, I found https://registry.terraform.io/providers/terraform-provider-graylog/graylog/latest/docs/guides/json-string-attribute so we don't seem to be the only ones doing this 😆

mitchnielsen: Thanks for the context. Thinking this over, too, and I'll post some links I'm still digesting. In the meantime though I'm happy to move forward with this and we can tweak as needed once we get more experience (and testing) in place.

https://discuss.hashicorp.com/t/how-to-handle-json-input-into-data-resource-with-multiple-types-with-the-terraform-sdk/22896/2:

    The typical approach today is, I think, what you already tried: define a string attribute, often with a name ending in _json to be explicit about it, and then decode that value as JSON inside the provider code.

This seems to confirm your string attribute thought if I'm reading it correctly.

https://github.com/hashicorp/terraform-plugin-go?tab=readme-ov-file#getting-started:

    When handling requests, [tfprotov5.DynamicValue](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-go/tfprotov5#DynamicValue) types should always have their [Unmarshal](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-go/tfprotov5#DynamicValue.Unmarshal) methods called; their properties should not be inspected directly. The [tftypes.Value](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-go/tfprotov5/tftypes#Value) returned from Unmarshal can be inspected to check [whether it is known](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-go/tfprotov5/tftypes#Value.IsKnown) and subsequently converted to a plain Go type using its [As](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-go/tfprotov5/tftypes#Value.As) method. As will return an error if the Value is not known.

Not sure how much crossover this has with the Terraform Dynamic Type docs you linked earlier, but this was mentioned in the first link I commented.

Small aside, I found https://registry.terraform.io/providers/terraform-provider-graylog/graylog/latest/docs/guides/json-string-attribute so we don't seem to be the only ones doing this 😆

mitchnielsen:
Was able to replicate the testing, as well as running `terraform plan` in between so it read the current state and reported no changes to make ✅

internal/api/block_documents.go
```diff
@@ -18,33 +18,28 @@ type BlockDocumentClient interface {
...
+	BlockTypeID   uuid.UUID              `json:"block_type_id"`
 }
 
 type BlockDocumentUpdate struct {
-	BlockSchemaID     *uuid.UUID `json:"block_schema_id"`
-	Data              string     `json:"data"`
-	MergeExistingData bool       `json:"merge_existing_data"`
+	BlockSchemaID     *uuid.UUID             `json:"block_schema_id"`
+	Data              map[string]interface{} `json:"data"`
+	MergeExistingData bool                   `json:"merge_existing_data"`
```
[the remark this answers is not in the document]
parkedwards: i think you're right, but the opposite - `merge=false` seems to be the overwriting option.  bc this is how we use the parameter in the update endpoint:

```python
        if merge_existing_data:
            # merge the existing data and the new data for partial updates
            current_data.update(update_values["data"])
            update_values["data"] = current_data
```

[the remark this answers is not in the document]
parkedwards: im down to remove it after you implement `Update()` and find it's ALWAYS going to be `false`

internal/provider/resources/block.go
```diff
@@ -116,16 +120,111 @@ func (r *BlockResource) Schema(_ context.Context, _ resource.SchemaRequest, resp
...
 				CustomType:  customtypes.UUIDType{},
-				Description: "Workspace ID (UUID) where the Block is located",
+				Description: "Workspace ID (UUID) where the Block is located. In Prefect Cloud, either the resource or the provider's `workspace_id` must be set in order to manage the Block.",
 			},
 		},
 	}
 }
 
+// copyBlockToModel maps an API response to a model that is saved in Terraform state.
+func copyBlockToModel(block *api.BlockDocument, state *BlockResourceModel) diag.Diagnostics {
```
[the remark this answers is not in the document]
parkedwards: that makes sense, bc the `Create()` method pulls from the `req.Plan`, whereas the `Read()` pulls from `req.State`

i'm down to call it a `model` in the helper (since it's shared - otherwise we'd have to call it something like `planOrStateModel` 😆 ) and then we can use the variable names you mentioned in the logic itself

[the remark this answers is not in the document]
parkedwards: this might be a more clear way to disambiguate in the helper

https://github.com/PrefectHQ/terraform-provider-prefect/pull/196/commits/26e2a761cd306912a49f2d8eafc8e300b4dddefa

internal/provider/resources/block.go
```diff
@@ -116,16 +120,117 @@ func (r *BlockResource) Schema(_ context.Context, _ resource.SchemaRequest, resp
...
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	blockTypeClient, err := r.client.BlockTypes(config.AccountID.ValueUUID(), config.WorkspaceID.ValueUUID())
+	if err != nil {
+		resp.Diagnostics.Append(helpers.CreateClientErrorDiagnostic("Block Types", err))
+
+		return
```
parkedwards, line 160: i got held up here bc i forgot to return after the client init errors.....kept running into weird nil pointer exceptions before i realized that i wasn't passing forward a non-nil client if a validation failed 🤦

mitchnielsen (approved):
Awesome work ✅

internal/provider/resources/block.go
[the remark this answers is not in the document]
mitchnielsen: I like it 👍🏼

PR #199, opened 2024-06-07
chore(hygiene): standardize model variable names + use planmodifiers for `created`

### LINKED ISSUE #195: Standardize on use of "plan" and "state" variables
## Summary

Context: https://github.com/PrefectHQ/terraform-provider-prefect/pull/193#discussion_r1626391852

Let's standardize on the use of `plan` and `state` for variable names when interacting with Terraform. For example:

```go

func (r *BlockResource) Read(ctx context.Context, req resource.ReadRequest, resp *resource.ReadResponse) {
	var state BlockResourceModel
        // ...
```

This is how the [Terraform provider tutorial](https://developer.hashicorp.com/terraform/tutorials/providers-plugin-framework/providers-plugin-framework-resource-update) names them, and (imo) is more intuitive to work with.

## Acceptance criteria
- [ ] Variable names are all set to `plan` and `state` instead of `model`, etc.

### LINKED ISSUE #200: hygiene: audit all computed resource schema attributes + set planmodifier for relatively static values
it should actually be for computed values that are immutable - `id` and `created`

since they aren't set by us, they are `computed` (and come from the API).  however, when we run `terraform plan`, the framework doesn't know whether or not those values WILL change, so we'll often find plan outputs that look like this:

```shell
  ~ update in-place

  # prefect_block.foo will be updated in-place
  ~ resource "prefect_block" "foo" {
      ~ created   = "2024-06-06T23:19:59Z" -> (known after apply)
      ~ data      = (sensitive value)
        id        = "626d4f5b-58e8-4604-adcf-d3415c40c1dd"
        name      = "foo"
      ~ updated   = "2024-06-06T23:20:17Z" -> (known after apply)
        # (1 unchanged attribute hidden)
    }

prefect_block.foo: Modifying... [id=626d4f5b-58e8-4604-adcf-d3415c40c1dd]
prefect_block.foo: Modifications complete after 0s [id=626d4f5b-58e8-4604-adcf-d3415c40c1dd]
```

however, `created` will never change after the resource is created.  similar to `id`, these are immutable values that ARENT set by us, and so we can use a planmodifier here to tell the framework to use the existing state as the baseline when running a plan 

https://developer.hashicorp.com/terraform/plugin/framework/resources/plan-modification#usestateforunknown

> [UseStateForUnknown()](https://developer.hashicorp.com/terraform/plugin/framework/resources/plan-modification#usestateforunknown): Copies the prior state value, if not null. This is useful for reducing (known after apply) plan outputs for computed attributes which are known to not change over time.

parkedwards:
resolves https://github.com/PrefectHQ/terraform-provider-prefect/issues/195
resolves https://github.com/PrefectHQ/terraform-provider-prefect/issues/200

high-level changes:
1. all `copy<Resource>ToModel()` helpers will use the `tfModel` variable name to disambiguate which model (plan, config, state) is being written to.  But still be clear about what it is
2. all resource model variables are renamed to be standardized, so that something like `req.State()` will be called `state`, `req.Plan()` will be a `plan`, and `req.Config()` will be a `config`
3. use a [UseStateForUnknown](https://developer.hashicorp.com/terraform/plugin/framework/resources/plan-modification#usestateforunknown) for `created` attributes, to reduce `(known after apply)` outputs for immutable attributes

### DIFF
--- a/internal/provider/resources/variable.go
+++ b/internal/provider/resources/variable.go
@@ -98,6 +98,9 @@ func (r *VariableResource) Schema(_ context.Context, _ resource.SchemaRequest, r
 				Computed:    true,
 				CustomType:  customtypes.TimestampType{},
 				Description: "Timestamp of when the resource was created (RFC3339)",
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.UseStateForUnknown(),
+				},
 			},
 			"updated": schema.StringAttribute{
 				Computed:    true,
--- a/internal/provider/resources/variable.go
+++ b/internal/provider/resources/variable.go
@@ -157,21 +157,21 @@ func copyVariableToModel(ctx context.Context, variable *api.Variable, tfModel *V
 
 // Create creates the resource and sets the initial Terraform state.
 func (r *VariableResource) Create(ctx context.Context, req resource.CreateRequest, resp *resource.CreateResponse) {
-	var config VariableResourceModel
+	var plan VariableResourceModel
 
 	// Populate the model from resource configuration and emit diagnostics on error
-	resp.Diagnostics.Append(req.Config.Get(ctx, &config)...)
+	resp.Diagnostics.Append(req.Plan.Get(ctx, &plan)...)
 	if resp.Diagnostics.HasError() {
 		return
 	}
 
 	var tags []string
-	resp.Diagnostics.Append(config.Tags.ElementsAs(ctx, &tags, false)...)
+	resp.Diagnostics.Append(plan.Tags.ElementsAs(ctx, &tags, false)...)
 	if resp.Diagnostics.HasError() {
 		return
 	}
 
-	client, err := r.client.Variables(config.AccountID.ValueUUID(), config.WorkspaceID.ValueUUID())
+	client, err := r.client.Variables(plan.AccountID.ValueUUID(), plan.WorkspaceID.ValueUUID())
 	if err != nil {
 		resp.Diagnostics.AddError(
 			"Error creating variable client",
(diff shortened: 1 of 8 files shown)


parkedwards:
supersedes #108

internal/provider/resources/variable.go
```diff
@@ -133,41 +136,42 @@ func (r *VariableResource) Schema(_ context.Context, _ resource.SchemaRequest, r
...
-	model.Tags = tags
+	tfModel.Tags = tags
 
 	return nil
 }
 
 // Create creates the resource and sets the initial Terraform state.
 func (r *VariableResource) Create(ctx context.Context, req resource.CreateRequest, resp *resource.CreateResponse) {
-	var model VariableResourceModel
+	var config VariableResourceModel
```
mitchnielsen, line 160: I don't mind using `config` at all, I just noticed that the [Terraform provider tutorial for Create](https://developer.hashicorp.com/terraform/tutorials/providers-plugin-framework/providers-plugin-framework-resource-create#implement-create-functionality) uses `plan` here. Which do you prefer?

mitchnielsen (approved):
Looks great - just one non-blocking question 👍🏼

internal/provider/resources/variable.go
parkedwards: it's a good question - im actually inclined to move these over to a `req.Plan` on the creates, since `req.Config` appears to be the raw, user-inputted object

```golang
// CreateRequest represents a request for the provider to create a
// resource. An instance of this request struct is supplied as an argument to
// the resource's Create function.
type CreateRequest struct {
	// Config is the configuration the user supplied for the resource.
	//
	// This configuration may contain unknown values if a user uses
	// interpolation or other functionality that would prevent Terraform
	// from knowing the value at request time.
	Config tfsdk.Config

	// Plan is the planned state for the resource.
	Plan tfsdk.Plan

	// ProviderMeta is metadata from the provider_meta block of the module.
	ProviderMeta tfsdk.Config
}
```

parkedwards: updated here: https://github.com/PrefectHQ/terraform-provider-prefect/pull/199/commits/7b8a1e1641f0785b38ab4619a2cb48daaf3eb929

PR #201, opened 2024-06-07
feat(flow): add Flow resource

skyscrapr:
Added support for `Flow` resource.

All acceptance tests are passing.

Test output:

```
Running tool: /opt/homebrew/bin/go test -timeout 30s -run ^TestAccResource_flow$ github.com/prefecthq/terraform-provider-prefect/internal/provider/resources

ok  	github.com/prefecthq/terraform-provider-prefect/internal/provider/resources	4.248s
```


docs/data-sources/account_members.md
```diff
@@ -17,7 +17,14 @@ Use this data source to obtain user or actor IDs to manage Workspace Access.
...
 ```terraform
-# Query all Members in Account
+terraform {
+  required_providers {
+    prefect = {
+      source = "prefecthq/prefect"
+    }
+  }
+}
+
```
mitchnielsen, line 27: The provider specification is implied, looks like we don't specify this in any of the other data source docs:

```
~/code/github.com/prefecthq/terraform-provider-prefect/main
$ rg 'required_providers' docs/data-sources

$
```

We can just remove it - what do you think?

examples/data-sources/prefect_account_members/data-source.tf
```diff
@@ -1,2 +1,9 @@
-# Query all Members in Account
+terraform {
+  required_providers {
+    prefect = {
+      source = "prefecthq/prefect"
+    }
+  }
+}
+
```
mitchnielsen, line 8: Same here, the provider config is implied, and we don't specify this in any of the other examples.

internal/provider/resources/flow.go
```diff
@@ -0,0 +1,372 @@
...
+		},
+	}
+}
+
+// copyFlowToModel copies an api.Flow to a FlowResourceModel.
+func copyFlowToModel(ctx context.Context, flow *api.Flow, model *FlowResourceModel) diag.Diagnostics {
+	model.ID = types.StringValue(flow.ID.String())
+	model.Created = customtypes.NewTimestampPointerValue(flow.Created)
+	model.Updated = customtypes.NewTimestampPointerValue(flow.Updated)
+	// model.WorkspaceID = customtypes.NewUUIDValue(flow.WorkspaceID)
```
mitchnielsen, line 140:
```suggestion
```

I believe we can remove this.

internal/provider/resources/flow.go
```diff
@@ -0,0 +1,372 @@
...
+	if resp.Diagnostics.HasError() {
+		return
+	}
+}
+
+// // ImportState imports the resource into Terraform state.
+// func (r *FlowResource) ImportState(ctx context.Context, req resource.ImportStateRequest, resp *resource.ImportStateResponse) {
+// 	resource.ImportStatePassthroughID(ctx, path.Root("id"), req, resp)
+// }
+
```
mitchnielsen, line 334:
```suggestion
```

internal/provider/resources/flow.go
```diff
@@ -0,0 +1,372 @@
...
+	resp.Diagnostics.Append(resp.State.Set(ctx, &model)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+}
+
+// Update updates the resource and sets the updated Terraform state on success.
+func (r *FlowResource) Update(ctx context.Context, req resource.UpdateRequest, resp *resource.UpdateResponse) {
+	// Unsupported
```
mitchnielsen, line 283: Any reason this is marked as unsupported? https://app.prefect.cloud/api/docs#tag/Flows/operation/update_flow_api_accounts__account_id__workspaces__workspace_id__flows__id__patch

internal/provider/resources/flow.go
```diff
@@ -0,0 +1,372 @@
...
+	// 	// The error from the API call should take precedence
+	// 	// followed by this custom error if a specific deployment is not returned
+	// 	if err == nil && len(deployments) != 1 {
+	// 		err = fmt.Errorf("a deployment with the handle=%s could not be found", model.Handle.ValueString())
+	// 	}
+
+	// 	if len(deployments) == 1 {
+	// 		deployment = deployments[0]
+	// 	}
+	// }
```
mitchnielsen, line 259: What's the plan for this part of the logic?

internal/provider/resources/flow.go
```diff
@@ -0,0 +1,372 @@
...
+	}
+
+	// if model.ID.IsNull() && model.Handle.IsNull() {
+	// 	resp.Diagnostics.AddError(
+	// 		"Both ID and Handle are unset",
+	// 		"This is a bug in the Terraform provider. Please report it to the maintainers.",
+	// 	)
+
+	// 	return
+	// }
```
mitchnielsen, line 218:
```suggestion
```

internal/testutils/provider.go
```diff
@@ -17,7 +17,7 @@ import (
 
 // TestAccPrefix is the prefix set for all resources created via acceptance testing,
 // so that we can easily identify and clean them up in case of flakiness/failures.
-const TestAccPrefix = "terraformacc"
+const TestAccPrefix = "tfacc"
```
mitchnielsen, line 20: Why the change here?

mitchnielsen (changes requested):
Thanks again for the contribution @skyscrapr 🤝 

I did a first pass review - just a couple suggestions and looks like there's a few instances of commented out logic that should be either uncommented or removed.

Let me know if I can clarify anything.

skyscrapr:
Yep. Happy to remove. 
Will update. 

Thanks for suggestion.

internal/provider/resources/flow.go
skyscrapr: no plan, there is no need for this code. removed.

internal/provider/resources/flow.go
skyscrapr: The default behaviour would be delete and recreate on a change. 
It's unclear if the use case to update in place adds much value.
Keen to understand if there's requirement.

internal/testutils/provider.go
skyscrapr: there is a limit to the length of this string

skyscrapr:
I had to regenerate docs due to the feedback and I've updated the branch with the suggested changes.

internal/provider/resources/flow.go
skyscrapr: My suggestion would be to see if customers need this capability. 
From what i can see the effort to update is not adding much.

internal/testutils/provider.go
skyscrapr: actually i just retried this and it appears to no longer be the case.
Ill revert this change

internal/provider/resources/flow.go
mitchnielsen: @parkedwards - could you weigh in here since you have more context? I'm thinking that if we choose not to support `Update` then let's remove the method entirely to avoid confusion while developing.

parkedwards: i think @skyscrapr's intuition is correct here. it looks like we only accept changes to the `tags` parameter on the flow resource. practically, we'd probably re-create the flow on the API if we _**did**_ need to make any material change

that said, i do think we need to keep `Update` in place, as we need to implement the `resource.Resource` interface

skyscrapr: Yes. The method has to remain. 
But the standard way to not support update is to leave this method blank. 

The core will then treat any changes in the config as a destroy/create for the resource.

mitchnielsen: Got it, thank you both 👍🏼

mitchnielsen: @skyscrapr - given the result of the discussion here, it sounds to me like we don't want users to be able to update Flows and instead force a replace. We can use PlanModifiers for that. For example:

```diff
diff --git a/internal/provider/resources/flow.go b/internal/provider/resources/flow.go
index d568c2b..669bae3 100644
--- a/internal/provider/resources/flow.go
+++ b/internal/provider/resources/flow.go
@@ -12,6 +12,7 @@ import (
        "github.com/hashicorp/terraform-plugin-framework/resource"
        "github.com/hashicorp/terraform-plugin-framework/resource/schema"
        "github.com/hashicorp/terraform-plugin-framework/resource/schema/listdefault"
+       "github.com/hashicorp/terraform-plugin-framework/resource/schema/listplanmodifier"
        "github.com/hashicorp/terraform-plugin-framework/resource/schema/planmodifier"
        "github.com/hashicorp/terraform-plugin-framework/resource/schema/stringplanmodifier"
        "github.com/hashicorp/terraform-plugin-framework/types"
@@ -127,6 +128,9 @@ func (r *FlowResource) Schema(_ context.Context, _ resource.SchemaRequest, resp
                                Optional:    true,
                                Computed:    true,
                                Default:     listdefault.StaticValue(defaultEmptyTagList),
+                               PlanModifiers: []planmodifier.List{
+                                       listplanmodifier.RequiresReplace(),
+                               },
                        },
                },
        }
```

Without this change, when I try to update an existing Flow resource with an additional tag, I get:

```

  # prefect_flow.flow will be updated in-place
  ~ resource "prefect_flow" "flow" {
      ~ created = "2024-06-13T19:29:06Z" -> (known after apply)
        id      = "54e09681-9535-4f0b-9a15-ffd999c7187e"
        name    = "test-flow"
      ~ tags    = [
            "test",
          + "test2",
        ]
      ~ updated = "2024-06-13T19:29:06Z" -> (known after apply)
    }

Do you want to perform these actions?
  Terraform will perform the actions described above.
  Only 'yes' will be accepted to approve.

  Enter a value: yes

prefect_flow.flow: Modifying... [id=54e09681-9535-4f0b-9a15-ffd999c7187e]
╷
│ Error: Provider produced inconsistent result after apply
│
│ When applying changes to prefect_flow.flow, provider "provider[\"registry.terraform.io/prefecthq/prefect\"]" produced an unexpected new value:
│ .tags: element 1 has vanished.
│
│ This is a bug in the provider, which should be reported in the provider's own issue tracker.
```

_With_ the diff applied using the PlanModifier, I now get something more helpful:

```
-/+ destroy and then create replacement

  # prefect_flow.flow must be replaced
-/+ resource "prefect_flow" "flow" {
      ~ created = "2024-06-13T19:29:06Z" -> (known after apply)
      ~ id      = "54e09681-9535-4f0b-9a15-ffd999c7187e" -> (known after apply)
        name    = "test-flow"
      ~ tags    = [ # forces replacement
            "test",
          + "test2",
        ]
      ~ updated = "2024-06-13T19:29:06Z" -> (known after apply)
    }

Plan: 1 to add, 0 to change, 1 to destroy.

Do you want to perform these actions?
  Terraform will perform the actions described above.
  Only 'yes' will be accepted to approve.

  Enter a value: yes

prefect_flow.flow: Destroying... [id=54e09681-9535-4f0b-9a15-ffd999c7187e]
prefect_flow.flow: Destruction complete after 0s
prefect_flow.flow: Creating...
prefect_flow.flow: Creation complete after 0s [id=c153db34-3aa9-4551-b07b-7e58df3ac9e6]

Apply complete! Resources: 1 added, 0 changed, 1 destroyed
```

What do you think?

skyscrapr: Looks good to me. I added a test to check for updating tags in the flow resource. I had to add the plan modifier suggested to get the correct functionality. tests pass so confirms that changing tags will work correctly.

good suggestion thanks!

internal/provider/resources/flow.go
```diff
@@ -0,0 +1,355 @@
...
+	client, err := r.client.Flows(model.AccountID.ValueUUID(), model.WorkspaceID.ValueUUID())
+	if err != nil {
+		resp.Diagnostics.AddError(
+			"Error creating deployment client",
+			fmt.Sprintf("Could not create deployment client, unexpected error: %s. This is a bug in the provider, please report this to the maintainers.", err.Error()),
+		)
+	}
+
+	// A deployment can be imported + read by either ID or Handle
+	// If both are set, we prefer the ID
```
mitchnielsen, line 231: based on the logic below it looks like we're only importing by ID for now, right?

internal/provider/resources/flow.go
```diff
@@ -0,0 +1,355 @@
...
+}
+
+// Update updates the resource and sets the updated Terraform state on success.
+func (r *FlowResource) Update(ctx context.Context, req resource.UpdateRequest, resp *resource.UpdateResponse) {
+	// Unsupported - redeploy by default
+}
+
+// Delete deletes the resource and removes the Terraform state on success.
+func (r *FlowResource) Delete(ctx context.Context, req resource.DeleteRequest, resp *resource.DeleteResponse) {
+	var model FlowResourceModel
```
mitchnielsen, line 276: Let's change this variable name to `state` per my earlier comment.

internal/provider/resources/flow.go
```diff
@@ -0,0 +1,355 @@
...
+
+	resp.Diagnostics.Append(resp.State.Set(ctx, &model)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+}
+
+// Read refreshes the Terraform state with the latest data.
+func (r *FlowResource) Read(ctx context.Context, req resource.ReadRequest, resp *resource.ReadResponse) {
+	var model FlowResourceModel
```
mitchnielsen, line 205: Let's change this variable name to `state` per my earlier comment.

internal/provider/resources/flow.go
```diff
@@ -0,0 +1,355 @@
...
+	if resp.Diagnostics.HasError() {
+		return
+	}
+}
+
+// ImportState imports the resource into Terraform state.
+func (r *FlowResource) ImportState(ctx context.Context, req resource.ImportStateRequest, resp *resource.ImportStateResponse) {
+	// we'll allow input values in the form of:
+	// - "workspace_id,name"
+	// - "name"
```
mitchnielsen, line 322: Could you please add a new directory showing examples for importing Flows?

For example, you can reference [examples/resources/prefect_variable/](https://github.com/skyscrapr/terraform-provider-prefect/tree/resource_flow/examples/resources/prefect_variable).

internal/client/flows.go
```diff
@@ -0,0 +1,196 @@
...
+	"net/http"
+
+	"github.com/google/uuid"
+
+	"github.com/prefecthq/terraform-provider-prefect/internal/api"
+)
+
+var _ = api.FlowsClient(&FlowsClient{})
+
+// FlowsClient is a client for working with Deployments.
```
mitchnielsen, line 18:
```suggestion
// FlowsClient is a client for working with Flows.
```

internal/provider/resources/flow.go
```diff
@@ -0,0 +1,355 @@
...
+	// 	)
+
+	// 	return
+	// }
+
+	client, err := r.client.Flows(model.AccountID.ValueUUID(), model.WorkspaceID.ValueUUID())
+	if err != nil {
+		resp.Diagnostics.AddError(
+			"Error creating deployment client",
+			fmt.Sprintf("Could not create deployment client, unexpected error: %s. This is a bug in the provider, please report this to the maintainers.", err.Error()),
```
mitchnielsen, line 226:
```suggestion
			"Error creating flow client",
			fmt.Sprintf("Could not create flow client, unexpected error: %s. This is a bug in the provider, please report this to the maintainers.", err.Error()),
```

internal/provider/resources/flow.go
```diff
@@ -0,0 +1,355 @@
...
+		)
+	}
+
+	// A deployment can be imported + read by either ID or Handle
```
mitchnielsen, line 230:
```suggestion
	// A flow can be imported + read by either ID or Handle
```

internal/provider/resources/flow_test.go
```diff
@@ -0,0 +1,65 @@
...
+	resourceName := "prefect_flow.flow"
+	workspaceResourceName := "prefect_workspace.workspace"
+	randomName := testutils.TestAccPrefix + acctest.RandStringFromCharSet(10, acctest.CharSetAlphaNum)
+
+	resource.ParallelTest(t, resource.TestCase{
+		ProtoV6ProviderFactories: testutils.TestAccProtoV6ProviderFactories,
+		PreCheck:                 func() { testutils.AccTestPreCheck(t) },
+		Steps: []resource.TestStep{
+			{
+				// Check creation + existence of the deployment resource
```
mitchnielsen, line 39:
```suggestion
				// Check creation + existence of the flow resource
```

internal/provider/resources/flow.go
mitchnielsen: Currently, when I try to import a flow, I get:

```
$ tf import prefect_flow.flow 85f06a57-d187-416e-b49a-8beb307b9afb
prefect_flow.flow: Importing from ID "85f06a57-d187-416e-b49a-8beb307b9afb"...
prefect_flow.flow: Import prepared!
  Prepared prefect_flow for import
prefect_flow.flow: Refreshing state...
╷
│ Error: Plugin did not respond
│
│ The plugin encountered an error, and failed to respond to the plugin6.(*GRPCProvider).ReadResource call. The plugin logs may contain more details.
╵


Stack trace from the terraform-provider-prefect plugin:

panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x0 pc=0x105491a50]

goroutine 66 [running]:
github.com/prefecthq/terraform-provider-prefect/internal/provider/resources.copyFlowToModel({0x1057aa0e8, 0x14000489560}, 0x0, 0x140000ee3c0)
        /Users/mitch/code/github.com/prefecthq/terraform-provider-prefect/resource_flow/internal/provider/resources/flow.go:141 +0x30
github.com/prefecthq/terraform-provider-prefect/internal/provider/resources.(*FlowResource).Read(0x1400048ae50, {0x1057aa0e8, 0x14000489560}, {{{{0x1057af758, 0x140005160f0}, {0x10569c100, 0x14000516030}}, {0x1057b11d8, 0x14000433040}}, 0x140004440c0, ...}, ...)
        /Users/mitch/code/github.com/prefecthq/terraform-provider-prefect/resource_flow/internal/provider/resources/flow.go:258 +0x39c
github.com/hashicorp/terraform-plugin-framework/internal/fwserver.(*Server).ReadResource(0x1400018d340, {0x1057aa0e8, 0x14000489560}, 0x140004895f0, 0x14000297690)
        /Users/mitch/go/pkg/mod/github.com/hashicorp/terraform-plugin-framework@v1.4.2/internal/fwserver/server_readresource.go:101 +0x510
github.com/hashicorp/terraform-plugin-framework/internal/proto6server.(*Server).ReadResource(0x1400018d340, {0x1057aa0e8?, 0x14000488e70?}, 0x140004a2600)
        /Users/mitch/go/pkg/mod/github.com/hashicorp/terraform-plugin-framework@v1.4.2/internal/proto6server/server_readresource.go:55 +0x2e0
github.com/hashicorp/terraform-plugin-go/tfprotov6/tf6server.(*server).ReadResource(0x1400038e000, {0x1057aa0e8?, 0x140004884e0?}, 0x14000530060)
        /Users/mitch/go/pkg/mod/github.com/hashicorp/terraform-plugin-go@v0.22.1/tfprotov6/tf6server/server.go:776 +0x3c8
github.com/hashicorp/terraform-plugin-go/tfprotov6/internal/tfplugin6._Provider_ReadResource_Handler({0x105778520, 0x1400038e000}, {0x1057aa0e8, 0x140004884e0}, 0x140001a4800, 0x0)
        /Users/mitch/go/pkg/mod/github.com/hashicorp/terraform-plugin-go@v0.22.1/tfprotov6/internal/tfplugin6/tfplugin6_grpc.pb.go:482 +0x1c0
google.golang.org/grpc.(*Server).processUnaryRPC(0x140001a1000, {0x1057aa0e8, 0x140004883f0}, {0x1057b0060, 0x140000bf860}, 0x1400021e000, 0x14000383050, 0x105d26688, 0x0)
        /Users/mitch/go/pkg/mod/google.golang.org/grpc@v1.62.1/server.go:1386 +0xb58
google.golang.org/grpc.(*Server).handleStream(0x140001a1000, {0x1057b0060, 0x140000bf860}, 0x1400021e000)
        /Users/mitch/go/pkg/mod/google.golang.org/grpc@v1.62.1/server.go:1797 +0xb10
google.golang.org/grpc.(*Server).serveStreams.func2.1()
        /Users/mitch/go/pkg/mod/google.golang.org/grpc@v1.62.1/server.go:1027 +0x8c
created by google.golang.org/grpc.(*Server).serveStreams.func2 in goroutine 13
        /Users/mitch/go/pkg/mod/google.golang.org/grpc@v1.62.1/server.go:1038 +0x13c

Error: The terraform-provider-prefect plugin crashed!

This is always indicative of a bug within the plugin. It would be immensely
helpful if you could report the crash with the plugin's maintainers so that it
can be fixed. The output above should help diagnose the issue.
```

mitchnielsen (changes requested):
Thanks for the changes @skyscrapr 🤝 

I did another pass through the code along with some testing and have some more small requests.

internal/provider/resources/flow.go
skyscrapr: hmm that's odd. the acceptance test for import passes fine. Let me take a closer look at this one.

skyscrapr: ok, some investigation into this. Import is working for the following scenario:
`terraform import prefect_flow.flow workspace_id,flow_id`
But in other scenarios it panics.
I'm looking into it.

internal/provider/resources/flow.go
```diff
@@ -0,0 +1,355 @@
...
+		return diags
+	}
+	model.Tags = tags
+
+	return nil
+}
+
+// Create creates the resource and sets the initial Terraform state.
+func (r *FlowResource) Create(ctx context.Context, req resource.CreateRequest, resp *resource.CreateResponse) {
+	var model FlowResourceModel
```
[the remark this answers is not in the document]
skyscrapr: done

internal/provider/resources/flow.go
skyscrapr: done

internal/provider/resources/flow.go
skyscrapr: ok, I made a couple of updates to support import correctly using ids. 
It's possible to do this via name too. 
But suggest we wait to see if any users request this as it meets the basic requirements currently.
Please take another look.

internal/provider/resources/flow.go
skyscrapr: changed to support ids only.

internal/provider/resources/flow.go
skyscrapr: done

internal/provider/resources/flow.go
mitchnielsen: Really appreciate the iteration here, thanks @skyscrapr.

Looks like we just need to better handle the scenario where the `inputParts` only equals `1`. You can see [block's import logic as an example](https://github.com/PrefectHQ/terraform-provider-prefect/blob/5ff52364630ea269fc336c8e66601a3cb381ad26/internal/provider/resources/block.go). I was able to copy some of that logic and get `tf import` working:

```diff
diff --git a/internal/provider/resources/flow.go b/internal/provider/resources/flow.go
index d26e379..97c609d 100644
--- a/internal/provider/resources/flow.go
+++ b/internal/provider/resources/flow.go
@@ -333,10 +333,17 @@ func (r *FlowResource) ImportState(ctx context.Context, req resource.ImportState
 		return
 	}
 
-	id := inputParts[maxInputCount-1]
-	if len(inputParts) == maxInputCount {
-		workspaceID := inputParts[0]
-		resp.Diagnostics.Append(resp.State.SetAttribute(ctx, path.Root("workspace_id"), workspaceID)...)
+	identifier := inputParts[0]
+	resp.Diagnostics.Append(resp.State.SetAttribute(ctx, path.Root("id"), identifier)...)
+
+	if len(inputParts) == 2 && inputParts[1] != "" {
+		workspaceID, err := uuid.Parse(inputParts[1])
+		if err != nil {
+			resp.Diagnostics.AddError("problem parsing workspace ID", "see details")
+
+			return
+		}
+
+		resp.Diagnostics.Append(resp.State.SetAttribute(ctx, path.Root("workspace_id"), workspaceID.String())...)
 	}
-	resp.Diagnostics.Append(resp.State.SetAttribute(ctx, path.Root("id"), id)...)
 }
```

```shell
$ tf import prefect_flow.flow c9bb05e4-2efe-49f7-b4e2-318d295553b6
prefect_flow.flow: Importing from ID "c9bb05e4-2efe-49f7-b4e2-318d295553b6"...
prefect_flow.flow: Import prepared!
  Prepared prefect_flow for import
prefect_flow.flow: Refreshing state... [id=c9bb05e4-2efe-49f7-b4e2-318d295553b6]

Import successful!

The resources that were imported are shown above. These resources are now in
your Terraform state and will henceforth be managed by Terraform.
```

>  But suggest we wait to see if any users request this as it meets the basic requirements currently.

I'm not opposed to that, but let's at least confirm that diagnostics properly return this information to communicate that workspace ID currently needs to be supplied (meaning this won't work on OSS for now). Currently, `id := inputParts[maxInputCount-1]` throws an error via stack trace rather than diagnostics. So let's either get that method working based of my example above, or throw correct diagnostics for now 👍🏼

skyscrapr: ok, applied suggested changes.

internal/provider/resources/flow.go
skyscrapr: updated as suggested

mitchnielsen (approved):
I'll be out for the next week and a half so I'll hand this over to @parkedwards for a final look.

Thanks again for the iteration on this one @skyscrapr 🤝 

Using the config below, I was able to create, update (replace), import, and destroy a Flow. Looking good from my side ✅ 

```terraform
terraform {
  required_providers {
    prefect = {
      source = "registry.terraform.io/prefecthq/prefect"
    }
  }
}

provider "prefect" {
  endpoint = "http://localhost:4200/api"
}

resource "prefect_flow" "flow" {
  name = "test-flow"
  tags = ["test", "test2", "test3"]
}
```

skyscrapr:
Awesome. Thanks for the thorough review. Once this is merged I'll update #202 to the same level. 

Cheers

internal/provider/resources/flow.go
```diff
@@ -0,0 +1,347 @@
...
+	resp.Diagnostics.Append(resp.State.Set(ctx, &state)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+}
+
+// ImportState imports the resource into Terraform state.
+func (r *FlowResource) ImportState(ctx context.Context, req resource.ImportStateRequest, resp *resource.ImportStateResponse) {
+	// we'll allow input values in the form of:
+	// - "workspace_id,id"
```
parkedwards, line 311:
```suggestion
	// - "id,workspace_id"
```

internal/provider/resources/flow.go
```diff
@@ -0,0 +1,347 @@
...
+	// - "id"
+	maxInputCount := 2
+	inputParts := strings.Split(req.ID, ",")
+
+	// eg. "foo,bar,baz"
+	if len(inputParts) > maxInputCount {
+		resp.Diagnostics.AddError(
+			"Unexpected Import Identifier",
+			fmt.Sprintf("Expected a maximum of 2 import identifiers, in the form of `workspace_id,name`. Got %q", req.ID),
```
parkedwards, line 320:
```suggestion
			fmt.Sprintf("Expected a maximum of 2 import identifiers, in the form of `name,workspace_id`. Got %q", req.ID),
```

internal/provider/resources/flow.go
```diff
@@ -0,0 +1,347 @@
...
+				// https://github.com/hashicorp/terraform-plugin-framework/issues/754
+				Description: "Flow ID (UUID)",
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.UseStateForUnknown(),
+				},
+			},
+			"created": schema.StringAttribute{
+				Computed:    true,
+				CustomType:  customtypes.TimestampType{},
+				Description: "Timestamp of when the resource was created (RFC3339)",
```
parkedwards, line 102: can we use a a `stringplanmodifier.UseStateForUnknown()` planmodifier for `created` here?

internal/provider/helpers/import.go
```diff
@@ -0,0 +1,26 @@
...
+
+import (
+	"fmt"
+
+	"github.com/google/uuid"
+	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+	"github.com/hashicorp/terraform-plugin-testing/terraform"
+)
+
+func GetResourceWorkspaceImportStateID(resourceName string, workspaceName string) resource.ImportStateIdFunc {
```
parkedwards, line 11: nice

internal/provider/resources/flow.go
skyscrapr: Interesting suggestion. Let me take a look at that to see what the experience is.

skyscrapr: since flows currently don't support updating, I'm not sure this adds anything to the overall experience. Happy either way.

parkedwards: ah that's right, they'll be re-created every time. we can disregard

parkedwards (approved):
LGTM - thank you!

Issue #175, opened 2024-05-23
Blocks Support: Create `prefect_block {}` datasource


parkedwards:
https://www.notion.so/prefect/Design-Block-support-in-the-Terraform-Provider-ae2ea25a18864de9b51b8aa06af9486b?pvs=4

the use case is probably more limited here - I'd imagine most people will want to create new Blocks rather than querying existing ones.  Maybe they'll want to set a Block Document's access level

But if for whatever reason, we need to, we should be able to query an existing Block Document:

```hcl
data "prefect_block" "my_existing_secret" {
  id = "00000000-0000-0000-0000-000000000000"
}

# or

data "prefect_block" "my_existing_secret" {
  name = "my-secret-name"
}
```

we dont have a "get block document by name" endpoint, but possibly might be able to use the filter endpoint?
https://app.prefect.cloud/api/docs#tag/Block-documents/operation/read_block_documents_api_accounts__account_id__workspaces__workspace_id__block_documents_filter_post

otherwise, we can just query a datasource by `id` and use the existing `GET /block_documents/{id}` endpoint

PR #207, opened 2024-06-10
Add 'Block' to the object support table

mitchnielsen:
Adds 'Block' to the object support table. We can mark 'Datasource support' as complete when https://github.com/PrefectHQ/terraform-provider-prefect/issues/175 is complete.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,19 +25,19 @@ We're continuously adding more Prefect Cloud object support, striving for near-p
 
 Check back with us to see new additions and improvements - and please don't hesitate to peruse our [Contributing section!](#contributing)
 
-| Prefect Cloud object | Datasource support? | Resource support? | Import support? |
-|----------------------:|:---------------------:|:-------------------:|:-----------------:|
-| Account Member       |       &check;       |                   |                 |
-| Account Role         |       &check;       |                   |                 |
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -30,6 +30,7 @@ Check back with us to see new additions and improvements - and please don't hesi
 | Account Member       |       &check;       |                   |                 |
 | Account Role         |       &check;       |                   |                 |
 | Account              |       &check;       |      &check;      |     &check;     |
+| Block                |                     |      &check;      |     &check;     |
```
mitchnielsen, line 33: I'm going to try to find a list of objects in Prefect, ideally one that delineates which are Cloud-only. We may want to make that distinction somewhere in our TF provider docs.

parkedwards: hah i was just thinking that. from this list, i think we can say:

Cloud-only:
- Account Member
- Account Role
- Account
- Service Account
- Team
- Workspace Access
- Workspace Role
- Workspace

Cloud and OSS
- Block
- Variable
- Work Pool

mitchnielsen: Sweet, thanks - updated in 4daaabc.

parkedwards (approved):

PR #209, opened 2024-06-10
feat(blocks): add Block data source

### LINKED ISSUE #175: Blocks Support: Create `prefect_block {}` datasource

mitchnielsen:
## Summary

Closes https://github.com/PrefectHQ/terraform-provider-prefect/issues/175

## To do
- [x] Consider adding fields to the schema that represent nested objects (see https://github.com/PrefectHQ/terraform-provider-prefect/pull/209#issuecomment-2159262200)
- [x] Consider getting block by `name` (see https://github.com/PrefectHQ/terraform-provider-prefect/pull/209#issuecomment-2159316724)
- [x] Consider adding tests (see https://github.com/PrefectHQ/terraform-provider-prefect/pull/209#issuecomment-2159332243)

## Testing

First, start up the local instance:

```shell
docker compose up -d
```

Create a `main.tf` file with these contents:

```terraform
terraform {
  required_providers {
    prefect = {
      source = "registry.terraform.io/prefecthq/prefect"
    }
  }
}

provider "prefect" {
  endpoint = "http://localhost:4200/api"
}

resource "prefect_block" "secret" {
  name      = "foo"
  type_slug = "secret"

  data = jsonencode({
    "value" : "bar"
  })
}

data "prefect_block" "my_existing_secret" {
  id = prefect_block.secret.id
}

output "block_info" {
  value     = data.prefect_block.my_existing_secret
  sensitive = true
}
```

Test it:

<details>
<summary>Click to expand</summary>

```
  + create
 <= read (data resources)

  # data.prefect_block.my_existing_secret will be read during apply
  # (config refers to values not yet known)
 <= data "prefect_block" "my_existing_secret" {
      + block_schema_id = (known after apply)
      + block_type_id   = (known after apply)
      + block_type_name = (known after apply)
      + created         = (known after apply)
      + data            = (sensitive value)
      + id              = (known after apply)
      + name            = (known after apply)
      + updated         = (known after apply)
    }

  # prefect_block.secret will be created
  + resource "prefect_block" "secret" {
      + created   = (known after apply)
      + data      = (sensitive value)
      + id        = (known after apply)
      + name      = "foo"
      + type_slug = "secret"
      + updated   = (known after apply)
    }

Plan: 1 to add, 0 to change, 0 to destroy.

Changes to Outputs:
  + block_info = (sensitive value)

Do you want to perform these actions?
  Terraform will perform the actions described above.
  Only 'yes' will be accepted to approve.

  Enter a value: yes

prefect_block.secret: Creating...
prefect_block.secret: Creation complete after 0s [id=b98844a2-39a5-4ba3-9562-0f104cc4c33f]
data.prefect_block.my_existing_secret: Reading...
data.prefect_block.my_existing_secret: Read complete after 0s [id=b98844a2-39a5-4ba3-9562-0f104cc4c33f]

Apply complete! Resources: 1 added, 0 changed, 0 destroyed.

Outputs:

block_info = <sensitive>
```

</details>

Get the specific value:

```
$ tf output -json block_info | jq .data
"{\"value\":\"bar\"}"
```


mitchnielsen:
Currently, some of the fields in the schema are omitted because I'm not yet sure how to represent types other than `string`, `UUID`, or `timestamp`. Here's the full response from the API [Read document by ID](https://app.prefect.cloud/api/docs#tag/Block-documents/operation/read_block_document_by_id_api_accounts__account_id__workspaces__workspace_id__block_documents__id__get):

<details>
<summary>Click to expand</summary>

```json
{
  "block_documents": {
    "operator": "and_",
    "id": {
      "any_": [
        "497f6eca-6276-4993-bfeb-53cbbbba6f08"
      ]
    },
    "is_anonymous": {
      "eq_": false
    },
    "block_type_id": {
      "any_": [
        "497f6eca-6276-4993-bfeb-53cbbbba6f08"
      ]
    },
    "name": {
      "any_": [
        "string"
      ],
      "like_": "my-block%"
    }
  },
  "block_types": {
    "name": {
      "like_": "marvin"
    },
    "slug": {
      "any_": [
        "string"
      ]
    }
  },
  "block_schemas": {
    "operator": "and_",
    "block_type_id": {
      "any_": [
        "497f6eca-6276-4993-bfeb-53cbbbba6f08"
      ]
    },
    "block_capabilities": {
      "all_": [
        "write-storage",
        "read-storage"
      ]
    },
    "id": {
      "any_": [
        "497f6eca-6276-4993-bfeb-53cbbbba6f08"
      ]
    },
    "version": {
      "any_": [
        "2.0.0",
        "2.1.0"
      ]
    }
  },
  "include_secrets": false,
  "sort": "NAME_ASC",
  "offset": 0,
  "limit": 0
}
```

</details>

So the current fields omitted are:

- `block_schema`
- `block_type`
- `block_document_references`

I'm also omitting `is_anonymous`, but I don't think we need to expose that (yet?).

-----

@parkedwards two quick questions:
- Do you see a need to add these currently-omitted fields to the data source?
- If yes, do you happen to know how to do it? I'm seeing some references to nested objects as a type but seems a little verbose.

mitchnielsen:
Looking into querying by name:

In the [Read Block Documents](https://app.prefect.cloud/api/docs#tag/Block-documents/operation/read_block_documents_api_accounts__account_id__workspaces__workspace_id__block_documents_filter_post) endpoint, we could allow a user to pass a `name` value and then interpolate it to fit into the `block_documents.name.like` or `block_documents.name.any` field:

image

This seems doable, but will require a more detailed implementation to do that translation. At the moment, I'm thinking that querying by ID should cover most use cases (especially since users would likely reference the object by the friendly `data.prefect_block.my_block.id` identifier).

What do you think @parkedwards?

mitchnielsen:
I drafted a test, but looking at existing Datasource tests it seems like they read existing objects in Prefect. Wondering which project we create these in, since I can't read the encrypted variables in the GitHub project once they're written.

In this case I'd need to go create a block named `my_block` before committing the test (following the example from the `variable_test.go`).

```go
package datasources_test

import (
	"fmt"
	"testing"

	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
	"github.com/prefecthq/terraform-provider-prefect/internal/testutils"
)

func fixtureAccBlockByID(name string) string {
	return fmt.Sprintf(`
data "prefect_block" "test" {
	id = prefect_block.%s.id
}
`, name)
}

//nolint:paralleltest // we use the resource.ParallelTest helper instead
func TestAccDatasource_block(t *testing.T) {
	datasourceName := "data.prefect_block.test"
	blockName := "my_block"
	blockValue := "{\"value\":\"bar\"}"

	resource.ParallelTest(t, resource.TestCase{
		ProtoV6ProviderFactories: testutils.TestAccProtoV6ProviderFactories,
		PreCheck:                 func() { testutils.AccTestPreCheck(t) },
		Steps: []resource.TestStep{
			{
				Config: fixtureAccBlockByID(blockName),
				Check: resource.ComposeAggregateTestCheckFunc(
					resource.TestCheckResourceAttrSet(datasourceName, "id"),
					resource.TestCheckResourceAttr(datasourceName, "name", blockName),
					resource.TestCheckResourceAttr(datasourceName, "data", blockValue),
				),
			},
		},
	})
}
```

internal/provider/datasources/block.go
```diff
@@ -0,0 +1,201 @@
...
+	Created     customtypes.TimestampValue `tfsdk:"created"`
+	Updated     customtypes.TimestampValue `tfsdk:"updated"`
+	AccountID   customtypes.UUIDValue      `tfsdk:"account_id"`
+	WorkspaceID customtypes.UUIDValue      `tfsdk:"workspace_id"`
+
+	Name          types.String          `tfsdk:"name"`
+	Data          jsontypes.Normalized  `tfsdk:"data"`
+	BlockSchemaID customtypes.UUIDValue `tfsdk:"block_schema_id"`
+	BlockTypeID   customtypes.UUIDValue `tfsdk:"block_type_id"`
+	BlockTypeName types.String          `tfsdk:"block_type_name"`
```
parkedwards, line 40: im wondering if we just omit these.  

my thought process was - we include attributes on the datasource model if we use them to query OR if they are part of the response + we anticipate a user needing to use them <somewhere>

for query attributes, i think `id` and possibly `name` are valid

for response attributes - functionally, i was thinking that someone only really needs to pull down an existing block if they need (1) the id, so that they can bind it to a `prefect_block_access` resource, or (2) the data itself, so they can pass it into a downstream TF resource

internal/provider/datasources/block.go
```diff
@@ -0,0 +1,201 @@
...
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	client, err := d.client.BlockDocuments(state.AccountID.ValueUUID(), state.WorkspaceID.ValueUUID())
+	if err != nil {
+		resp.Diagnostics.AddError(
+			"Error creating block client",
+			fmt.Sprintf("Could not create block client, unexpected error: %s. This is a bug in the provider, please report this to the maintainers.", err.Error()),
+		)
```
parkedwards, line 139: for client init errors, i have a helper called `CreateClientErrorDiagnostic` that you can use

parkedwards:
> Do you see a need to add these currently-omitted fields to the data source?
If yes, do you happen to know how to do it? I'm seeing some references to nested objects as a type but seems a little verbose.

@mitchnielsen you're right, i think we can omit those, and i added a couple more in the comments above that we might think about skipping.  the nested object typing is a little nasty, and if we dont believe users to need to actually access those property values (which im not sure they do as of right now - that could change, at which point we can always take on the incremental work to support those), then we can omit

parkedwards:
> This seems doable, but will require a more detailed implementation to do that translation. At the moment, I'm thinking that querying by ID should cover most use cases (especially since users would likely reference the object by the friendly data.prefect_block.my_block.id identifier).

great idea, and totally do-able. i think we should do this (in this PR or a followup).  the general idea would be - allow the user to input either an `id` or a `name`, and based on the presence of either (you'd have to give precedence to one of those, so it'd probably be `id`) - you'd fire off a request to either the `Get(id uuid.UUID)` client method that we already have OR we can add a `List(names []string)` method to the client (similar to what you did with `BlockSchema` client)

check out the `variable.go` datasource, we do something similar there

https://github.com/PrefectHQ/terraform-provider-prefect/blob/main/internal/provider/datasources/variable.go#L121-L164

parkedwards:
> I drafted a test, but looking at existing Datasource tests it seems like they read existing objects in Prefect. Wondering which project we create these in, since I can't read the encrypted variables in the GitHub project once they're written.

@mitchnielsen yeah that was an approach i started with but dont love.  maybe we can create a new block resource in the same fixture string (right before your datasource) and then you can compare values against your input resource and the queried datasource

internal/provider/datasources/block.go
mitchnielsen: Ah right, thanks - updated in fc16a27. Was following along with the `Variable` example. We'll go back at some point and make use of that helper more consistently.

parkedwards: for sure. as you can tell, i am an abomination to the consistency gods

mitchnielsen: Hey you wrote the helper for consistency 😄

internal/provider/datasources/block.go
mitchnielsen: Sounds good to me, updated in d6b60ed.

mitchnielsen:
> > I drafted a test, but looking at existing Datasource tests it seems like they read existing objects in Prefect. Wondering which project we create these in, since I can't read the encrypted variables in the GitHub project once they're written.
> 
> @mitchnielsen yeah that was an approach i started with back then but dont love. maybe we can create a new block resource in the same fixture string (right before your datasource) and then you can compare values against your input resource and the queried datasource

Works for me. I was originally wondering if it should be cleaned up after the test run, but it's probably alright for now that it stays around since even concurrent tests won't try to modify it.

internal/provider/datasources/block_test.go
```diff
@@ -0,0 +1,73 @@
...
+  data = jsonencode({
+    "someKey" : "someValue"
+  })
+
+  account_id = "%s"
+  workspace_id = data.prefect_workspace.evergreen.id
+}
+
+data "prefect_block" "my_existing_secret_by_id" {
+  id = prefect_block.%s.id
```
parkedwards, line 31: you might need the `workspace_id` here too

mitchnielsen: TIL, thanks - that got me further. Now seeing:

```
    block_test.go:57: Step 1/2 error: Error running pre-apply plan: exit status 1
        
        Error: Error refreshing block state
        
          with data.prefect_block.my_existing_secret_by_name,
          on terraform_plugin_test.tf line 35, in data "prefect_block" "my_existing_secret_by_name":
          35: data "prefect_block" "my_existing_secret_by_name" {
        
        Could not read block, unexpected error: status code 404 Not Found,
        error={"detail":"Block document not found"}
```

That portion saying `running pre-apply plan` makes me think the test didn't get a chance to actually apply the plan yet, so expected that the resource won't be there.

mitchnielsen: Fixed by specifying `depends_on` in the datasources.

internal/api/block_documents.go
```diff
@@ -8,6 +8,7 @@ import (
 
 type BlockDocumentClient interface {
 	Get(ctx context.Context, id uuid.UUID) (*BlockDocument, error)
+	GetByName(ctx context.Context, typeSlug, name string) (*BlockDocument, error)
```
parkedwards, line 11: tiny nit - `GetByTypeSlug`

parkedwards: nvm, i misread this

mitchnielsen: Could do `GetBySlugAndName` but felt ugly 😆

parkedwards (approved):
nice

internal/provider/datasources/block.go
```diff
@@ -0,0 +1,198 @@
...
+type BlockDataSourceModel struct {
+	ID          customtypes.UUIDValue      `tfsdk:"id"`
+	Created     customtypes.TimestampValue `tfsdk:"created"`
+	Updated     customtypes.TimestampValue `tfsdk:"updated"`
+	AccountID   customtypes.UUIDValue      `tfsdk:"account_id"`
+	WorkspaceID customtypes.UUIDValue      `tfsdk:"workspace_id"`
+
+	Name          types.String         `tfsdk:"name"`
+	Data          jsontypes.Normalized `tfsdk:"data"`
+	BlockTypeName types.String         `tfsdk:"block_type_name"`
```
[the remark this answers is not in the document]
mitchnielsen: @parkedwards - could you do a quick sanity check here? Want to make sure I didn't miss anything - could just be inconsistency in the API field names.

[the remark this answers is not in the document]
parkedwards: yeah - so the choice i made (and it's open for debate) was that - since `prefect_block` is itself kind of a unification of multiple API resources, i figured we could do the same and just obfuscate the notion of a `block_type` and just call it a `type` on the resource, since we're already abstracting those underlying APIs to the user

and then the `_slug` aspect was to differentiate a type's `name` and a `slug` - in general, they are similar, but the `slug` is the hyphenated string, whereas i _**think**_ the name is cap-cased and has spaces?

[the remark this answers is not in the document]
parkedwards:
```
        "name": "Azure Container Instance Job",
        "slug": "azure-container-instance-job",
```

internal/provider/datasources/block.go
```diff
@@ -0,0 +1,198 @@
...
+		return
+	}
+
+	var block *api.BlockDocument
+
+	switch {
+	case !state.ID.IsNull():
+		block, err = client.Get(ctx, state.ID.ValueUUID())
+	case !state.Name.IsNull() && !state.BlockTypeName.IsNull():
+		block, err = client.GetByName(ctx, state.BlockTypeName.ValueString(), state.Name.ValueString())
```
parkedwards, line 136: yeah now that you mention it - i might actually modify the `BlockTypeName / block_type_name` to be `TypeSlug / type_slug`

the API expects a slug

```
@router.get(
    "/slug/{slug}/block_documents/name/{block_document_name}",
)
```

mitchnielsen: I'm down, the one caveat I see is that the [API docs](https://app.prefect.cloud/api/docs#tag/Block-documents/operation/read_block_document_by_id_api_accounts__account_id__workspaces__workspace_id__block_documents__id__get) specifically call it `block_type_name` 🤔

mitchnielsen: Mm, I see what you mean. I was mapping the [BlockDocument class from Nebula](https://github.com/PrefectHQ/nebula/blob/22e66a334369d96a08fc19a877c9117e94f388b8/src/prefect/server/schemas/core.py#L681-L704) to the [BlockDocument](https://github.com/PrefectHQ/terraform-provider-prefect/blob/b1f0f16ebd55d272d7f8aca29b76038b8576ad4d/internal/api/block_documents.go#L20-L31) struct, which is where I got `block_type_name`. Looks like I just need to mentally delinate when objects represent the API request vs the API response.

mitchnielsen: Iterated on this in 95ae8f5, lemme know what you think.

parkedwards (approved):
dude. awesome!

Issue #203, opened 2024-06-07
Reduce repetition in codebase


mitchnielsen:
As we build out the provider, we've started seeing opportunities to DRY up the codebase a bit. Let's use this top-level issue to track places we can reduce repetition in the codebase to keep it as consistent and maintainable as possible.

- [ ] Client creation in `internal/provider/resources/block.go` ([context](https://github.com/PrefectHQ/terraform-provider-prefect/pull/198#discussion_r1630416318))
- [ ] HTTP requests in `internal/client/*` (lots of repetition here when creating the request, setting headers, executing the request, and checking the status code; remember to pass the close function as a result so we can defer the call to it from the caller)
- [ ] Diagnostics creation (https://github.com/PrefectHQ/terraform-provider-prefect/issues/212)
- [ ] Enums for resource type name references (context: https://github.com/PrefectHQ/terraform-provider-prefect/pull/213#discussion_r1643587946)
- [ ] others?

PR #213, opened 2024-06-13
Improve Diagnostics consistency

### LINKED ISSUE #212: Use helper functions for consistent Diagnostics
## Summary

https://developer.hashicorp.com/terraform/plugin/framework/diagnostics#consistent-diagnostic-creation

Let's use existing helper functions to ensure consistent Diagnostic entries in the provider. We can also write new helper functions where needed for more consistency and less repetition.

Note that some diagnostic entries may not need to be replaced with helper functions, specifically those that are single-use and not repeated.

## Acceptance criteria
- [x] Diagnostic entries use helper functions for consistency (where it makes sense)

mitchnielsen:
## Summary

Closes https://github.com/PrefectHQ/terraform-provider-prefect/issues/212

- Uses helper functions more consistently
- Adds new helper functions
- Ensure `return` on diagnostic errors
- Fix diagnostic resource object references

## To do
- [x] Replace all repetitive Diagnostics with helper functions

### DIFF
--- a/internal/provider/datasources/account_member.go
+++ b/internal/provider/datasources/account_member.go
@@ -124,10 +124,7 @@ func (d *AccountMemberDataSource) Configure(_ context.Context, req datasource.Co
 
 	client, ok := req.ProviderData.(api.PrefectClient)
 	if !ok {
-		resp.Diagnostics.AddError(
-			"Unexpected Data Source Configure Type",
-			fmt.Sprintf("Expected api.PrefectClient, got: %T. Please report this issue to the provider developers.", req.ProviderData),
-		)
+		resp.Diagnostics.Append(helpers.ConfigureTypeErrorDiagnostic("data source", req.ProviderData))
 
 		return
 	}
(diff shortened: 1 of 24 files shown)


mitchnielsen:
This is a bigger one, but if you look at each commit, the changes are pretty repetitive/predictable. Also, a majority of the changes are lines removed rather than changed/added.

internal/provider/datasources/account_member.go
```diff
@@ -158,10 +155,9 @@ func (d *AccountMemberDataSource) Read(ctx context.Context, req datasource.ReadR
 	// workspaceRoles, err := client.List(ctx, []string{model.Name.ValueString()})
 	accountMembers, err := client.List(ctx, []string{config.Email.ValueString()})
 	if err != nil {
-		resp.Diagnostics.AddError(
-			"Error refreshing Account Member state",
-			fmt.Sprintf("Could not search for Account Members, unexpected error: %s", err.Error()),
-		)
+		resp.Diagnostics.Append(helpers.ResourceClientErrorDiagnostic("Account Member", "list", err))
```
parkedwards, line 158: not blocking, but wonder if we should make enums for the magical strings that we use for the resource type name

parkedwards (approved):

internal/provider/datasources/account_member.go
mitchnielsen: I think you're right, crossed my mind while working on this. Added to https://github.com/PrefectHQ/terraform-provider-prefect/issues/203.

PR #214, opened 2024-06-14
fix(blocks): add slug name to error messaging for block type/schema fetch failure

parkedwards:
when an incorrect slug is provided, the error can be hard to track down.  to help, we'll print the slug itself in the error message

this is the current error message
```
Could not get_by_slug Block Type, unexpected error: All attempts fail:
#1: status code 404 Not Found, error={"detail":"Block type not found"}
```


mitchnielsen (approved):
Love a good detailed error messsage

Issue #231, opened 2024-07-16
Replace usage of deprecated `xattr.TypeWithValidate` interface


mitchnielsen:
## Summary

In https://github.com/hashicorp/terraform-plugin-framework/releases/tag/v1.8.0, there's a note about a deprecated interface:

> attr/xattr: The TypeWithValidate interface has been deprecated in preference of the ValidateableAttribute interface. A ValidatableParameter interface has also been added to the function package (https://github.com/hashicorp/terraform-plugin-framework/issues/968)

We use this in two places:

1. https://github.com/PrefectHQ/terraform-provider-prefect/blob/37502a647233631ef85c34f811367f255349dbc5/internal/provider/customtypes/timestamp_type.go#L18
2. https://github.com/PrefectHQ/terraform-provider-prefect/blob/37502a647233631ef85c34f811367f255349dbc5/internal/provider/customtypes/uuid_type.go#L18

In https://developer.hashicorp.com/terraform/plugin/framework/validation#defining-value-validation, it discusses this:

> Value validation should be used in preference to Type validation. Refer to [Value Validation](https://developer.hashicorp.com/terraform/plugin/framework/validation#value-validation) for more information.

It might make sense to simply remove the type validation and keep the value validation we already have, but will need to confirm.

## Acceptance criteria
- [x] `TypeWithValidate` is no longer used (it's deprecated in 1.8.0, and not yet known when it will be removed)

PR #236, opened 2024-07-17
feat: add `Owner` as an option when setting an account role

### LINKED ISSUE #235: Support setting `Owner` as an account level role

jamiezieziula:
Resolves #235


internal/provider/resources/service_account.go
```diff
@@ -131,7 +131,7 @@ func (r *ServiceAccountResource) Schema(_ context.Context, _ resource.SchemaRequ
 				Description: "Account Role name of the service account",
```
mitchnielsen, line 131: Wonder if it's worth mentioning the possible values here:

1. It would tell you the list of possible values.
2. You can use a correct value the first time, instead of putting a wrong one and having it error out later.

It'd show up here:

image

mitchnielsen (approved):
Looks great, was able to:

- Create a new SA with "Owner" role
- Import an existing SA with "Owner" role
- Change the SA role to "Member" and back to "Owner"
- Destroy the SA

I also approved the CI acceptance tests run and all looks good.

Nice work 🚀

internal/provider/resources/service_account.go
mitchnielsen:
```suggestion
				Description: "Account Role name of the service account (one of: Admin, Member, Owner)",
```

Something like that!

examples/data-sources/prefect_account_role/data-source.tf
```diff
@@ -1,9 +1,14 @@
...
 
 # Read down the default Member Account Role
 data "prefect_account_role" "member" {
   name = "Member"
 }
+
+# Read down the default Owner Account Role
+data "prefect_account_role" "owner" {
+  name = "Owner"
+}
```
parkedwards, line 14: is pre-commit running by any chance

internal/provider/datasources/account_role.go
```diff
@@ -75,7 +75,7 @@ Use this data source read down the pre-defined Roles, to manage User and Service
 				Required:    true,
 				Description: "Name of the Account Role",
 				Validators: []validator.String{
-					stringvalidator.OneOf("Admin", "Member"),
+					stringvalidator.OneOf("Admin", "Member", "Owner"),
```
parkedwards, line 78: nice

parkedwards (approved):

examples/data-sources/prefect_account_role/data-source.tf
jamiezieziula: yes, but im getting an odd failure on this `internal/provider/customtypes/timestamp_type.go:18:6:`

mitchnielsen: Yeah that's probably https://github.com/PrefectHQ/terraform-provider-prefect/issues/231, it's just a deprecation and not a removal but for some reason `pre-commit` treats it as a failure. I'll work on a PR to update that.

Issue #241, opened 2024-08-14
Error in acceptance tests: "Error during create Workspace: 503 Service Unavailable"


mitchnielsen:
## Summary

We've been seeing consistent failures in acceptance tests:

```
Run make testacc
TF_ACC=1 make test
make[1]: Entering directory '/home/runner/work/terraform-provider-prefect/terraform-provider-prefect'
gotestsum --max-fails=50 ./...
∅  .
∅  internal/api
∅  internal/client
∅  internal/provider
∅  internal/provider/customtypes
∅  internal/provider/helpers
∅  internal/testutils
∅  internal/utils
✓  internal/provider/datasources (5.17[8](https://github.com/PrefectHQ/terraform-provider-prefect/actions/runs/10379690891/job/28738291855?pr=202#step:7:9)s)
✖  internal/provider/resources (10.815s)
=== Failed
=== FAIL: internal/provider/resources TestAccResource_flow (0.44s)
    flow_test.go:34: Step 1/3 error: Error running apply: exit status 1
        
        Error: Error during create Workspace
        
          with prefect_workspace.workspace,
          on terraform_plugin_test.tf line 12, in resource "prefect_workspace" "workspace":
          12: resource "prefect_workspace" "workspace" {
        
        Could not create Workspace, unexpected error: status code 503 Service
        Unavailable, error={"exception_message":"Service Unavailable"}
DONE 23 tests, 1 failure in 12.766s
make[1]: *** [Makefile:3[9](https://github.com/PrefectHQ/terraform-provider-prefect/actions/runs/10379690891/job/28738291855?pr=202#step:7:10): test] Error 1
make[1]: Leaving directory '/home/runner/work/terraform-provider-prefect/terraform-provider-prefect'
make: *** [Makefile:45: testacc] Error 2
```

Examples:
- Failure in `flow` test: https://github.com/PrefectHQ/terraform-provider-prefect/actions/runs/10379690891/job/28738291855?pr=202
- Failure in `workspace` test: https://github.com/PrefectHQ/terraform-provider-prefect/actions/runs/10377092449/job/28730515383

[...]

PR #242, opened 2024-08-14
Retry Workspace creation to mitigate 503s

mitchnielsen:
Mitigates the 503 service unavailable errors by retrying Workspace creation. Further information provided in comments in the code for posterity.

Related to https://github.com/PrefectHQ/terraform-provider-prefect/issues/241

## Testing

Run the Acceptance Tests repeatedly:
- Old behavior: resource tests would sometimes fail to create Workspaces with a `503 Service Unavailable error`
- New behavior: resource tests pass every time

### DIFF
--- a/internal/client/workspaces.go
+++ b/internal/client/workspaces.go
@@ -57,7 +57,12 @@ func (c *WorkspacesClient) Create(ctx context.Context, data api.WorkspaceCreate)
 	}
 	defer resp.Body.Close()
 
-	if resp.StatusCode != http.StatusCreated {
+	// Create is considered successful if the response is either:
+	// 1. StatusCreated (201) - meaning it was created
+	// 2. StatusConflict (409) - meaning it already exists.
+	//   This is relevant in cases where the Create method is retried.
+	// If the error is not one of these, then Create is considered unsuccessful.
+	if resp.StatusCode != http.StatusCreated && resp.StatusCode != http.StatusConflict {
 		errorBody, _ := io.ReadAll(resp.Body)
 
 		return nil, fmt.Errorf("status code %s, error=%s", resp.Status, errorBody)
--- a/internal/client/workspaces.go
+++ b/internal/client/workspaces.go
@@ -57,12 +57,7 @@ func (c *WorkspacesClient) Create(ctx context.Context, data api.WorkspaceCreate)
 	}
 	defer resp.Body.Close()
 
-	// Create is considered successful if the response is either:
-	// 1. StatusCreated (201) - meaning it was created
-	// 2. StatusConflict (409) - meaning it already exists.
-	//   This is relevant in cases where the Create method is retried.
-	// If the error is not one of these, then Create is considered unsuccessful.
-	if resp.StatusCode != http.StatusCreated && resp.StatusCode != http.StatusConflict {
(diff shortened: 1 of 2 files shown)


jimid27 (approved):
This Lgtm but I'm still no go expert

jamiezieziula (approved):

internal/provider/resources/workspace.go
```diff
@@ -154,11 +156,25 @@ func (r *WorkspaceResource) Create(ctx context.Context, req resource.CreateReque
...
 
-	workspace, err := client.Create(ctx, api.WorkspaceCreate{
-		Name:        plan.Name.ValueString(),
-		Handle:      plan.Handle.ValueString(),
-		Description: plan.Description.ValueStringPointer(),
-	})
+	// In certain scenarios, such as in tests when multiple Workspaces are created at the same time
+	// for test isolation, Prefect will return a "503 Service Unavailable" error.
+	// To mitigate this, we will retry the Workspace creation. The Create method will be considered
+	// successful if either the Workspace was created or if the Workspace already exists.
```
parkedwards, line 162:
```suggestion
	// To mitigate this, we will retry the Workspace creation. The Create method will stop retrying
	// if either the Workspace was created or if the Workspace already exists.
```

internal/client/workspaces.go
```diff
@@ -57,7 +57,12 @@ func (c *WorkspacesClient) Create(ctx context.Context, data api.WorkspaceCreate)
 	}
 	defer resp.Body.Close()
 
-	if resp.StatusCode != http.StatusCreated {
+	// Create is considered successful if the response is either:
+	// 1. StatusCreated (201) - meaning it was created
+	// 2. StatusConflict (409) - meaning it already exists.
+	//   This is relevant in cases where the Create method is retried.
+	// If the error is not one of these, then Create is considered unsuccessful.
+	if resp.StatusCode != http.StatusCreated && resp.StatusCode != http.StatusConflict {
```
parkedwards, line 65: so if we get a 409, that means that the workspace exists already right

if we skip the early exit here on a 409, what's the expected behavior

parkedwards: basically, im wondering if we should still exit on a 409 early, and find some way to pass signal back to the retry method to cease attempts (which i think is what we want to do)

parkedwards: bc if we proceed here on a 409, the error will be some nil pointer exception from a failed json unmarshalling, i think

parkedwards: maybe: https://github.com/avast/retry-go?tab=readme-ov-file#func--attemptsforerror

mitchnielsen: I think I see what you mean - what you're suggesting would be more explicit and stable. Let me try 👍🏼

mitchnielsen: @parkedwards - I started down that path but ran into some blockers/questions, so I pushed 834de5f34 to simplify this a bit. My reasoning is in the commit message - if you disagree I'm happy to try that option again assuming we can address the blockers/questions 👍🏼

parkedwards (approved):

internal/client/workspaces.go
parkedwards: all good, i think the important part is retrying to begin with

PR #245, opened 2024-08-20
feat(Deployments): support Parameters

mitchnielsen:
## Summary

Supports setting Parameters on Deployment resouces.

Related to https://github.com/PrefectHQ/terraform-provider-prefect/issues/238

## Testing

First, see that the acceptance tests passed ✅ 

Also tested manually:

```terraform
terraform {
  required_providers {
    prefect = {
      source = "registry.terraform.io/prefecthq/prefect"
    }
  }
}

provider "prefect" {
  endpoint = "https://api.stg.prefect.dev"

  account_id   = "my-account-id"
  workspace_id = "<my-workspace-id>"

  api_key = "<my-api-key>"
}

resource "prefect_workspace" "workspace" {
  handle = "mitch-testing"
  name   = "mitch-testing"
}

resource "prefect_flow" "flow" {
  name         = "my-flow"
  workspace_id = prefect_workspace.workspace.id
  tags         = ["tf-test"]
}

resource "prefect_deployment" "deployment" {
  name                     = "my-deployment"
  description              = "string"
  workspace_id             = prefect_workspace.workspace.id
  flow_id                  = prefect_flow.flow.id
  entrypoint               = "hello_world.py:hello_world"
  tags                     = ["test"]
  enforce_parameter_schema = false
  manifest_path            = "./bar/foo"
  parameters               = jsonencode({ "some-parameter" : "some-value" })
  path                     = "./foo/bar"
  paused                   = false
  version                  = "v1.1.1"
  work_pool_name           = "mitch-testing-pool"
  work_queue_name          = "default"
}
```

```
$ go install
...

$ tf apply
...
```

image

Then I updated my config to add another field with an `int` and applied it, and ran a job:

image

Issue #238, opened 2024-07-18
Add support for Deployments


mitchnielsen:
## Summary

As a follow-up to support for Flows ([https://github.com/PrefectHQ/terraform-provider-prefect/issues/5](https://github.com/PrefectHQ/terraform-provider-prefect/issues/5), [https://github.com/PrefectHQ/terraform-provider-prefect/pull/202](https://github.com/PrefectHQ/terraform-provider-prefect/pull/202)) let's add support for Deployments.

## References

* Docs:[ deployment schema](https://docs.prefect.io/3.0/deploy/index#deployment-schema)

## To do

- [X] Initial Deployments support: [https://github.com/PrefectHQ/terraform-provider-prefect/pull/202](https://github.com/PrefectHQ/terraform-provider-prefect/pull/202)
- [ ] Add support for remaining [fields](https://app.prefect.cloud/api/docs#tag/Deployments/operation/create_deployment_api_accounts__account_id__workspaces__workspace_id__deployments__post)
  - [X] `parameter_openapi_schema`: [https://github.com/PrefectHQ/terraform-provider-prefect/pull/263](https://github.com/PrefectHQ/terraform-provider-prefect/pull/263)
  - [X] `parameters` -> [https://github.com/PrefectHQ/terraform-provider-prefect/pull/245](https://github.com/PrefectHQ/terraform-provider-prefect/pull/245)
  - [X] `storage_document_id`: [https://github.com/PrefectHQ/terraform-provider-prefect/pull/262](https://github.com/PrefectHQ/terraform-provider-prefect/pull/262)
  - [X] `job_variables` -> [https://github.com/PrefectHQ/terraform-provider-prefect/pull/250](https://github.com/PrefectHQ/terraform-provider-prefect/pull/250)
  - [X] Sub-issues

felixpelletier:
If I may cast my vote: the most important params for my use case would be:
- parameter_openapi_schema (for convenience, since the HTTP PATCH method for that particular param doesn't work in the REST API)
- Deployment Access (by far the most important to us, but we can manage by using a local-exec provisioner on a null_resource and apply it on every terraform apply. No big deal.)
- schedules

mitchnielsen:
Thanks @felixpelletier, this helps - I updated the description to add `access` to the top and moved `parameter_openapi_schema` and `schedules` up.

PR #250, opened 2024-08-26
feat(Deployments): support Job Variables

felixpelletier:
## Summary
Supports setting Job Variables on Deployment resources.

Related to https://github.com/PrefectHQ/terraform-provider-prefect/issues/238

Note: I'm a Go newbie, but I mainly just followed the very similar PR here: https://github.com/PrefectHQ/terraform-provider-prefect/pull/245

## Testing
I ran the unit tests, but I don't have an appropriate setup to run the acceptance tests. 
If someone could help with that, I'd be very grateful 😉


mitchnielsen:
This is great, thanks for your contribution @felixpelletier! Will take a look 👍🏼

mitchnielsen:
Hi @felixpelletier, I pushed one small change 309ab47. I found that `env` was a key that works for all work pool types, so I wanted to use that in the example and tests to confirm that your changes work as expected (and they do 👍🏼).

mitchnielsen (approved):
This looks great, thanks again @felixpelletier 🎉 

- New field added in correct places
- Tests added and passing
- Docs updated

Will approve and hand to @parkedwards for a final lookover since I ended up pushing a couple small changes.

parkedwards (approved):

Issue #254, opened 2024-09-05
Changing Variable to a JSON value in the UI makes next Terraform run fail


felixpelletier:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request.
* Please do not leave _+1_ or _me too_ comments, they generate extra noise for issue followers and do not help prioritize the request.
* If you are interested in working on this issue or have submitted a pull request, please leave a comment.

### Terraform Version

terraform 0.15.3 (I know it's old, but I don't believe the issue is there)
provider v2.2.2

### Affected Resource(s)

* prefect_variable

### Terraform Configuration Files

```tf
resource "prefect_variable" "some_variable" {
  name = "some_variable_name"
  value = jsonencode([])
}
```

### Expected Behavior

After creating the variable, I expected the variable's value to be a JSON.

After modifying the value from `"[]"` to `["some_value"]` (notice the absence of surrounding double-quotes) in the UI and trying to re-apply the terraform plan: I would expect it to convert the JSON to a String so I can decode it and use it.

### Actual Behavior

After creating the variable, the value is a string `"[]"`

Applying the plan results in an error:
`Could not get Variable, unexpected error: failed to decode response: json: cannot unmarshal array into Go struct field Variable.value of type string`

Even when adding an `ignore_changes` lifecycle rule, the problem persists. (Which makes sense. It's still a data source).

### Steps to Reproduce

1. `terraform apply`
2. In the Cloud UI, change the value from `"[]"` to `["some_value"]`
3. `terraform apply`

### Important Factoids

N/A

### References

N/A

parkedwards:
thanks for raising @felixpelletier - let us take a look at this

mitchnielsen:
Alrighty, looks like I was able to replicate this:

First, I created a variable in the Prefect UI and gave it a value of `["some_value"]`.

Then I created the Terraform config to read that variable as a datasource:

```terraform
```

mitchnielsen:
So what's the fix? We'll want to change the underlying Terraform Type of Variable values from `string` to `JSON`. This is due to changes made fairly recently to Variables in the API to make them JSON objects instead of just strings (example: https://github.com/PrefectHQ/prefect/pull/13543).

[...]

PR #277, opened 2024-10-04
fix(variable): change 'value' type to Dynamic

mitchnielsen:
## Summary

Changes the Variable 'value' type from string to Dynamic.

In https://github.com/PrefectHQ/prefect/pull/13543 and associated changes around May 2024, Variables were updated from simple strings to JSON objects. The Terraform provider has still been treating them as strings, so when folks tried to put JSON-compatible values in them, Terraform would fail to work with them as found in https://github.com/PrefectHQ/terraform-provider-prefect/issues/254

That said, this doesn't mean you can always expect a full JSON object in response. It could be individual JSON-compatible types, like `bool` or `int`. So we actually need a Dynamic attribute type here.

Related to https://linear.app/prefect/issue/PLA-247/changing-variable-to-a-json-value-in-the-ui-makes-next-terraform-run

Related to https://github.com/PrefectHQ/terraform-provider-prefect/issues/254

## References

- https://developer.hashicorp.com/terraform/language/expressions/types
- https://developer.hashicorp.com/terraform/language/expressions/type-constraints
- https://developer.hashicorp.com/terraform/plugin/framework/handling-data/dynamic-data
- https://developer.hashicorp.com/terraform/plugin/framework/handling-data/attributes/dynamic

From https://docs.prefect.io/3.0/develop/variables:

> (Variables) are named, mutable values of any JSON type

Also, you can see the accepted types for the `value` field here:

```
$ curl -s https://api.prefect.cloud/api/openapi.json | jq '.components.schemas.Variable.properties.value.anyOf'
$ curl -s https://api.prefect.cloud/api/openapi.json | jq '.components.schemas.Variable.properties.value.anyOf[]'
{
  "type": "string",
  "maxLength": 5000
}
{
  "type": "integer"
}
{
  "type": "number"
}
{
  "type": "boolean"
}
{
  "type": "object"
}
{
  "items": {},
  "type": "array"
}
```


mitchnielsen:
So changing this to a JSON type may not be the full fix here. As the API response shows, the value here can be almost anything: string, integer, number, boolean, object, and array. So if I put `5` in the Variable, the provider will fail to convert that into `map[string]interface`. Need to think about this some more.

mitchnielsen:
We may need to use a dynamic attribute here: https://developer.hashicorp.com/terraform/plugin/framework/handling-data/attributes/dynamic

docs/resources/variable.md
```diff
@@ -25,7 +25,7 @@ resource "prefect_variable" "example" {
 ### Required
 
 - `name` (String) Name of the variable
-- `value` (String) Value of the variable
+- `value` (Dynamic) Value of the variable, supported Terraform value types: string, number, bool, tuple, object
```
parkedwards, line 28: brutal!

internal/provider/datasources/variable.go
```diff
@@ -182,3 +194,60 @@ func (d *VariableDataSource) Read(ctx context.Context, req datasource.ReadReques
 		return
 	}
 }
+
+// getDynamicValue converts the 'value' attribute from a native Go type to a DynamicValue.
+func getDynamicValue(model VariableDataSourceModel, variable *api.Variable) (basetypes.DynamicValue, diag.Diagnostics) {
```
parkedwards, line 199: v nice

parkedwards (approved):

PR #279, opened 2024-10-07
Adds support for cloud-run-v2, modal work pool metadata

### LINKED ISSUE #276: Support cloud_run_v2 for the worker_metadata data source
The `worker_metadata` datasource supports `cloud-run` and `cloud-run:push` but not the equivalent `cloud-run-v2` and `cloud-run-v2:push`. We should support `v2` as well.

### Version

```
Terraform v1.9.7
on darwin_arm64
+ provider registry.terraform.io/prefecthq/prefect v2.2.3
```

EmilRex:
Closes #276


mitchnielsen:
Thanks so much for the PR @EmilRex! I'll take a look at this today, and in the meantime will kick off the pending tests.

mitchnielsen:
@EmilRex Taking a look into how this works behind the scenes. In the meantime, I wonder if these base job configs aren't set up / available in this environment? These tests use the staging instance in the `github-ci-tests` account.

EmilRex:
Ah, that might be it... I only checked against production to confirm that `cloud-run-v2` and `cloud-run-v2:push` were there.

mitchnielsen:
Alrighty, should be all set here @EmilRex. Thanks to a tip from @jakekaplan, I updated the endpoint to use based on if the target is OSS or Cloud, which should more closely match what the Prefect client is doing. Tests are now passing and I believe we've enumerated every type there is based on the docs (added a link to those in a comment for future reference).

mitchnielsen (approved):

PR #286, opened 2024-10-22
Add retries to all resources and datasources

mitchnielsen:
Uses the [hashicorp/retryablehttp](https://pkg.go.dev/github.com/hashicorp/go-retryablehttp#section-readme) package to get an HTTP client with
retries and exponential backoff out of the box. This is implemented
within our client package, meaning calls to the CRUD methods from both
resources and datasources will inherit that functionality.

At a high level, the changes here are:
- Replacing the `http` client with a `retryablehttp` client
- Reverting implementations of `avast/retry` in resources
- Removing `avast/retry`-related helpers

Related to https://linear.app/prefect/issue/PLA-183/add-retries-to-resources


mitchnielsen:
It probably makes sense just to implement retries on all resources/datasources at once, rather than incrementally adding them where we get reports of failures - I'm going to see if I can get Cursor to do most of the heavy lifting there, and if we can add a facility to make it easier to add/tweak in the future 🤔

parkedwards:
@mitchnielsen trying to think, are there calls that we WOULDNT want to retry on?

mitchnielsen:
> trying to think, are there calls that we WOULDNT want to retry on?

Asking in Slack 👌🏼

internal/client/client.go
```diff
@@ -10,14 +11,38 @@ import (
...
+	retryableClient.CheckRetry = checkRetryPolicy
+
+	// Finally, convert the retryablehttp client to a standard http client.
+	// This allows us to retain the `http.Client` interface, and avoid specifying
+	// the `retryablehttp.Client` interface in our client methods.
+	httpClient := retryableClient.StandardClient()
+
 	client := &Client{
-		hc: http.DefaultClient,
+		hc: httpClient,
```
mitchnielsen, line 45: This is the main change here - we're replacing the default HTTP client with our retry-able HTTP client.

internal/client/client.go
```diff
@@ -36,26 +61,6 @@ func New(opts ...Option) (*Client, error) {
...
-// WithClient configures the underlying http.Client used to send
-// requests.
-func WithClient(httpClient *http.Client) Option {
-	return func(client *Client) error {
-		client.hc = httpClient
-
-		return nil
-	}
-}
-
```
mitchnielsen, line 58: These were unused.

internal/client/client.go
```diff
@@ -97,3 +102,21 @@ func WithDefaults(accountID uuid.UUID, workspaceID uuid.UUID) Option {
...
+	// If the response is a 404 (NotFound), try again. This is particularly
+	// relevant for block-related objects that are created asynchronously.
+	if resp.StatusCode == http.StatusNotFound {
+		return true, err
+	}
+
+	// Fall back to the default retry policy for any other status codes.
+	//nolint:wrapcheck // we've extended this method, no need to wrap error
+	return retryablehttp.DefaultRetryPolicy(ctx, resp, err)
+}
```
mitchnielsen, line 122: Here we can be specific about what warrants a retry. This is cool because we knew in the past we didn't need to retry 409s, but the retry implementation we had at the time didn't really let us achieve that.

internal/client/client.go
```diff
@@ -10,14 +11,38 @@ import (
...
+
+	// By default, retryablehttp will only retry requests if there was some kind
+	// of transient server or networking error. We can be more specific with this
+	// by providing a custom function for determining whether or not to retry.
+	retryableClient.CheckRetry = checkRetryPolicy
+
+	// Finally, convert the retryablehttp client to a standard http client.
+	// This allows us to retain the `http.Client` interface, and avoid specifying
+	// the `retryablehttp.Client` interface in our client methods.
+	httpClient := retryableClient.StandardClient()
```
mitchnielsen, line 42: This part is pretty sweet. It saves a lot of lines changed - see https://github.com/PrefectHQ/terraform-provider-prefect/pull/286/commits/f86254b42555e37a760580ecda01a4705b52d009.

parkedwards (approved):
awesome!

PR #288, opened 2024-10-28
Docs: remove "Cloud" from resources that are not cloud-only

mitchnielsen:
Changes "Prefect Cloud" to "Prefect" for resources that are not cloud-only.

Context: reported in a [community Slack thread](https://prefect-community.slack.com/archives/CL09KU1K7/p1724127210012189) that Variables appeared to be Prefect Cloud only.

Related to https://linear.app/prefect/issue/PLA-452/cycle-5-catch-all


parkedwards:
@mitchnielsen should we close this branch?

mitchnielsen:
Ah thanks for the callout @parkedwards, didn't notice `terraform docs` had overwritten the changes. I pushed 8f6c724 so the changes are made in the source code, and then generated the docs changes.

parkedwards (approved):

PR #293, opened 2024-10-31
feat(automations): API client and data model for request/response

parkedwards:
resolves https://linear.app/prefect/issue/PLA-575/tfp-automations-set-up-api-client

In this PR, we'll set up the Automations API client

As a part of this work, we need to take a stab at handling the polymorphic schemas for `trigger` and `actions`.  In #305, I have some work drafted to handle this at the Terraform layer in a more detailed way.  For example, for `triggers`, we may have each individual schema defined in the Terraform resource + use a `oneOf()` validator to allow for validating each one

However, I don't believe we'll need to be as thorough at this API layer because it's just translating data to/from JSON.  I'm taking an approach here of just _**combining**_ all possible attributes across a polymorphic peer group and making the attributes optional.  I'm hoping that we can just key off of the `type` attribute and handle the translation at the resource level...but that may definitely change

I still needed to created some custom types here, like for `ResourceSpecification` and `Actions` in general.  Since there's some funky custom JSON marshalling logic, I threw in some unit tests for sanity

### DIFF
--- a/internal/api/automations.go
+++ b/internal/api/automations.go
@@ -23,6 +23,8 @@ type Automation struct {
 	AccountID   uuid.UUID `json:"account_id"`
 	WorkspaceID uuid.UUID `json:"workspace_id"`
 }
+
+// AutomationUpsert is the data needed to create or update an automation.
 type AutomationUpsert struct {
 	Name             string   `json:"name"`
 	Description      string   `json:"description"`
@@ -33,6 +35,11 @@ type AutomationUpsert struct {
 	ActionsOnResolve []Action `json:"actions_on_resolve"`
 }
 
+// Trigger defines the triggering conditions on an Automation.
+// On the API, a Trigger is a polymorphic type and can be represented
+// by several schemas based on the `type` attribute.
+// Here, we'll combine all possible attributes for each type
+// and make them optional.
 type Trigger struct {
 	Type string `json:"type"`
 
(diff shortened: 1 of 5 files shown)


internal/api/automations.go
```diff
@@ -0,0 +1,169 @@
...
+	MatchRelated *ResourceSpecification `json:"match_related,omitempty"`
+	Posture      *string                `json:"posture,omitempty"`
+	After        []string               `json:"after,omitempty"`
+	Expect       []string               `json:"expect,omitempty"`
+	ForEach      []string               `json:"for_each,omitempty"`
+	Threshold    *int                   `json:"threshold,omitempty"`
+	Within       *string                `json:"within,omitempty"` // Duration string
+	// For MetricTrigger
+	Metric *MetricTriggerQuery `json:"metric,omitempty"`
+	// For CompoundTrigger
```
parkedwards, line 57:
```suggestion
	// For CompoundTrigger and SequenceTrigger
```

internal/api/automations.go
```diff
@@ -0,0 +1,169 @@
...
+
+	// For EventTrigger
+	Match        *ResourceSpecification `json:"match,omitempty"`
+	MatchRelated *ResourceSpecification `json:"match_related,omitempty"`
+	Posture      *string                `json:"posture,omitempty"`
+	After        []string               `json:"after,omitempty"`
+	Expect       []string               `json:"expect,omitempty"`
+	ForEach      []string               `json:"for_each,omitempty"`
+	Threshold    *int                   `json:"threshold,omitempty"`
+	Within       *string                `json:"within,omitempty"` // Duration string
```
mitchnielsen, line 54: is this saying "Within" is similar to a duration? almost looks like this was going to be another struct field that lost its new line but wanted to check

internal/api/automations_test.go
```diff
@@ -0,0 +1,259 @@
...
+	"github.com/stretchr/testify/assert"
+	"k8s.io/utils/ptr"
+)
+
+func TestResourceSpecificationModel(t *testing.T) {
+	t.Parallel()
+	tests := []struct {
+		name string
+		json string
+		want api.ResourceSpecification
```
mitchnielsen, line 18: nice 👌🏼

mitchnielsen (approved):
looks good, this is a tough one

internal/api/automations.go
parkedwards: yeah the former - it's a string that reflects a duration

PR #294, opened 2024-10-31
feat(Deployments): support access control

mitchnielsen:
## Summary

Adds support for access control on Deployments.

Closes https://linear.app/prefect/issue/PLA-439/deployments-support-access

## Testing

- See the tests added
- See the `example.tf` file added


internal/provider/resources/deployment_access_test.go
```diff
@@ -0,0 +1,212 @@
+package resources_test
```
mitchnielsen, line 1: We don't currently have any tests for block access, which is the closest API to this one, so this may not be anywhere near perfect - but once we tune it we should be able to add some tests for block access according to these.

internal/api/deployments.go
```diff
@@ -82,40 +82,3 @@ type DeploymentUpdate struct {
...
-	RunActors    []Actor `json:"run_actors"`
-	ViewActors   []Actor `json:"view_actors"`
-}
-
-type Actor struct {
-	ID    uuid.UUID `json:"id"`
-	Name  string    `json:"name"`
-	Email string    `json:"email"`
-	Type  string    `json:"type"`
-}
```
mitchnielsen, line 121: Moved this to a separate file for clarity. We don't have a consistent pattern for this yet - for example, Workspaces have this more separated while Blocks have it together (in some places) - open to personal opinions

parkedwards: good call

examples/resources/prefect_deployment_access/resource.tf
```diff
@@ -0,0 +1,68 @@
...
+  deployment_id = prefect_deployment.test.id
+
+  manage_actor_ids = [prefect_service_account.test.actor_id]
+  run_actor_ids    = [prefect_service_account.test.actor_id]
+  view_actor_ids   = [prefect_service_account.test.actor_id]
+
+  manage_team_ids = [data.prefect_team.test.id]
+  run_team_ids    = [data.prefect_team.test.id]
+  view_team_ids   = [data.prefect_team.test.id]
+}
```
parkedwards, line 68: does this need a `terraform fmt --recursive ./`

internal/api/deployments_access.go
```diff
@@ -0,0 +1,44 @@
...
+	RunTeamIDs     []string `json:"run_team_ids"`
+	ViewTeamIDs    []string `json:"view_team_ids"`
+}
+
+// DeploymentAccessControl is a definition of deployment access control.
+type DeploymentAccessControl struct {
+	ManageActors []ObjectActorAccess `json:"manage_actors"`
+	RunActors    []ObjectActorAccess `json:"run_actors"`
+	ViewActors   []ObjectActorAccess `json:"view_actors"`
+}
```
parkedwards, line 44: i guess this is pretty similar to the Blocks access API then

internal/client/deployment_access.go
```diff
@@ -0,0 +1,100 @@
...
+func (c *Client) DeploymentAccess(accountID uuid.UUID, workspaceID uuid.UUID) (api.DeploymentAccessClient, error) {
+	if accountID == uuid.Nil {
+		accountID = c.defaultAccountID
+	}
+
+	if workspaceID == uuid.Nil {
+		workspaceID = c.defaultWorkspaceID
+	}
+
+	return &DeploymentAccessClient{
```
parkedwards, line 35: wonder if we should have an `IsCloudEndpoint` check and early exit here if no account/workspace ID is provided.  unless we support Deployment ACLs in the OSS server?

https://github.com/PrefectHQ/terraform-provider-prefect/blob/47f66ffd9e4afff2556b663c74a7c8ee2fe45880/internal/client/block_documents.go#L37-L39

internal/client/deployment_access.go
```diff
@@ -0,0 +1,100 @@
...
+		hc:          c.hc,
+		routePrefix: getWorkspaceScopedURL(c.endpoint, accountID, workspaceID, "deployments"),
+		apiKey:      c.apiKey,
+	}, nil
+}
+
+func (c *DeploymentAccessClient) Read(ctx context.Context, deploymentID uuid.UUID) (*api.DeploymentAccessControl, error) {
+	url := fmt.Sprintf("%s/%s/access", c.routePrefix, deploymentID.String())
```
parkedwards, line 43: do we need a trailing slash at the endpoint

internal/provider/resources/deployment_access.go
```diff
@@ -0,0 +1,349 @@
...
+	}
+
+	r.client = client
+}
+
+func (r *DeploymentAccessResource) Schema(_ context.Context, _ resource.SchemaRequest, resp *resource.SchemaResponse) {
+	defaultEmptyList, _ := basetypes.NewListValue(types.StringType, []attr.Value{})
+
+	resp.Schema = schema.Schema{
+		Description: "The resource `deployment_access` represents a connection between an accessor " +
```
parkedwards, line 69: i love using the markdown ```` ticks, but it does prevent us from writing more sane multiline strings here in the go layer

parkedwards (approved):
nice work

internal/provider/resources/deployment_access.go
mitchnielsen: Yeah I'm with you there. Could define them elsewhere in the file so they're more "left-aligned" 🤔

internal/client/deployment_access.go
mitchnielsen: Ah yes good catch, let me add that - to my knowledge this cloud-only because OSS won't have any concept of a workspace which is a requirement for ACLs to work

internal/api/deployments_access.go
mitchnielsen: Yeah very similar, glad you gave me some code to reference lol

examples/resources/prefect_deployment_access/resource.tf
mitchnielsen: Hmm that doesn't generate any changes for me, what are you seeing?

internal/client/deployment_access.go
mitchnielsen: -> cabc564

examples/resources/prefect_deployment_access/resource.tf
parkedwards: oh i just saw the missing newline, but i couldn't remember if terraform fmt handles that for us, or some other pre-commit

parkedwards (approved):

examples/resources/prefect_deployment_access/resource.tf
mitchnielsen: ahh the missing newline at the very end,  yeah looks like some example files have that and some don't, and our linters/formatters don't seem to be doing that for us

PR #299, opened 2024-11-04
Extend implementation of ephemeral workspaces

mitchnielsen:
## Summary

Related to https://linear.app/prefect/issue/PLA-188/testing-replace-evergreen-workspace-with-ephemeral-workspaces

Closes https://linear.app/prefect/issue/PLA-640/use-ephemeral-workspaces-in-remaining-tests

Change descriptions are available in the commit messages.


mitchnielsen:
This is just about done - all that's left is Workspace and Workspace_Access datasource tests. In a previous iteration we had a simple retry on Workspace creation, but that was removed when we added the `retryablehttp` package. The problem now is that in the tests, if I create an ephemeral workspace and then immediately test the datasource, it'll claim the workspace doesn't exist yet - when in reality it just needs some time to return it.

I'm working on implementing the retries, but this is going a bit beyond the scope of this change since it's starting to make some provider-level changes.

mitchnielsen:
Okay, the retry-related changes have been reverted and we'll address those in [PLA-641](https://linear.app/prefect/issue/PLA-641/use-ephemeral-workspaces-in-resources-with-race-conditions). That will keep this PR focused on updating tests.

jamiezieziula (approved):

PR #300, opened 2024-11-11
chore/Add infrastructure document id support

jimid27:
- Resolves: https://linear.app/prefect/issue/PLA-442/deployments-support-infrastructure-document-id
- Adds support and tests for `infrastructure_document_ids` in deployment specs
- I believe that the k8s job block is about as simple as I can make an infra block to pull the id from but happy for suggestions


jimid27:
Closing as this field has been deprecated.

jimid27:
> @jimid27 - do you happen to know if there's a timeline for this field to be removed entirely? If it's not particularly soon, I'm open to still merging this. In certain cases it could make it easier for folks to adopt the Terraform provider.
> 
> cc @parkedwards

It sounds like Cloud still supports this for previous versions of Prefect being used but it is all but removed in OSS per Jake

PR #301, opened 2024-11-12
Finish ephemeral workspace implementation, address race conditions

mitchnielsen:
Closes https://linear.app/prefect/issue/PLA-641/use-ephemeral-workspaces-in-resources-with-race-conditions

- Adds retry functionality on block schema retrieval, addressing the race conditions that can occur when creating ephemeral workspaces for block and deployment acceptance tests
- Implements ephemeral workspaces for the remaining resources and datasources

This means we no longer use the evergreen workspace, but we'll address removal and cleanup in https://linear.app/prefect/issue/PLA-642/ensure-ephemeral-workspaces-are-cleaned-up-consistently.

### DIFF
--- a/internal/provider/datasources/workspace_test.go
+++ b/internal/provider/datasources/workspace_test.go
@@ -8,46 +8,49 @@ import (
 	"github.com/prefecthq/terraform-provider-prefect/internal/testutils"
 )
 
-func fixtureAccWorkspaceByHandle(handle string) string {
+func fixtureAccWorkspaceByHandle(workspace, handle string) string {
 	return fmt.Sprintf(`
-data "prefect_workspace" "evergreen" {
+%s
+
+data "prefect_workspace" "testdata" {
 	handle = "%s"
+	depends_on = [prefect_workspace.test]
 }
-`, handle)
+`, workspace, handle)
 }
-func fixtureAccWorkspaceByID(id string) string {
+
+func fixtureAccWorkspaceByID(workspace string) string {
 	return fmt.Sprintf(`
-data "prefect_workspace" "evergreen" {
-	id = "%s"
+%s
+
+data "prefect_workspace" "testdata" {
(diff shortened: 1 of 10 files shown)


internal/provider/resources/block.go
```diff
@@ -130,6 +131,70 @@ func (r *BlockResource) Schema(_ context.Context, _ resource.SchemaRequest, resp
...
+// If no block schemas are returned, the retrieval is retried because Prefect creates
+// them asynchronously after the creation of a workspace.
+//
+//nolint:ireturn // required by Terraform API
+func (r *BlockResource) getLatestBlockSchema(ctx context.Context, plan BlockResourceModel) (*api.BlockSchema, diag.Diagnostic) {
+	var blockSchemas []*api.BlockSchema
+	var latestBlockSchema *api.BlockSchema
+	var diags diag.Diagnostic
+
+	err := retry.Do(func() error {
```
mitchnielsen, line 171: This is the special case in this PR where we need to retry block schemas because they're created asynchronously upon workspace creation.

For other resources, like workspace datasources, we can use`depends_on` to ensure a resource is created and ready prior to referencing it.

internal/provider/datasources/workspace_test.go
```diff
@@ -8,46 +8,49 @@ import (
...
 
-func fixtureAccWorkspaceByHandle(handle string) string {
+func fixtureAccWorkspaceByHandle(workspace, handle string) string {
 	return fmt.Sprintf(`
-data "prefect_workspace" "evergreen" {
+%s
+
+data "prefect_workspace" "testdata" {
 	handle = "%s"
+	depends_on = [prefect_workspace.test]
```
mitchnielsen, line 17: This addresses the race condition where the data source wouldn't be able to find the newly-created ephemeral workspace.

parkedwards (approved):
amazing!

PR #304, opened 2024-11-20
Deployments: support `schedule`

mitchnielsen:
## Summary

Closes https://linear.app/prefect/issue/PLA-440/deployments-support-schedules

## References
- https://docs.prefect.io/v3/deploy/infrastructure-concepts/prefect-yaml#schedule-fields
- https://docs.prefect.io/v3/automate/add-schedules#create-schedules-in-yaml
- https://docs.prefect.io/v3/api-ref/rest-api/server/deployments/read-deployment-schedules
- https://app.prefect.cloud/api/docs#tag/Deployments/operation/read_deployment_schedules_api_accounts__account_id__workspaces__workspace_id__deployments__id__schedules_get

## Notes

- The API offers the ability to manage schedules via the Deployments endpoints, but there's also endpoints specifically for Deployment Schedules. This would let us make Deployment Schedules a separate resource, but it may not be ideal since there's no endpoint for getting a specific Schedule - which we'd need to refresh state for a single Deployment Schedule resource in Terraform.
  - I ended up going with this separate endpoint and separate resource, and we can test multiple schedules in https://linear.app/prefect/issue/PLA-772/deployments-support-multiple-schedules. Schedules on Deployments was causing problems - see [this gist](https://gist.github.com/mitchnielsen/a3bae3a165f619f668c81047c6e97a4e).
- [Here is the gist](https://gist.github.com/mitchnielsen/c7306ca2cbdb3b34bef731cf12f2cb3d) where I tested schedules on the Deployment resource.


internal/provider/resources/deployment_schedule.go
```diff
@@ -0,0 +1,382 @@
...
+}
+
+//nolint:ireturn // required to return a diagnostic
+func validateSchedules(schedules []*api.DeploymentSchedule) diag.Diagnostic {
+	if len(schedules) != 1 {
+		return diag.NewErrorDiagnostic("Unsupported number of schedules", fmt.Sprintf("Expected 1 schedule, got %d. Only one schedule is supported.", len(schedules)))
+	}
+
+	return nil
+}
```
mitchnielsen, line 382: We'll have to limit to 1 schedule at a time for now. Opened https://linear.app/prefect/issue/PLA-772/deployments-support-multiple-schedules to investigate support for multiple schedules.

mitchnielsen: Added a note to the docs as well

parkedwards (approved):

PR #306, opened 2024-11-21
feat(deployments): add concurrency support

jimid27:
- Resolves: https://linear.app/prefect/issue/PLA-701/deployments-support-concurrency
- Adds concurrency limit which is an int and concurrency options which is an object
- Based on the current api docs for deployments found [here](https://docs.prefect.io/v3/api-ref/rest-api/server/deployments/create-deployment)


mitchnielsen:
🤔 Looking into concurrency_limit more. Looks like it is deprecated in favor of concurrency_limit_id based on [this code](https://github.com/PrefectHQ/prefect/blob/7903f46743f9a55439c85e5ad78d2ebd88ef869e/src/prefect/client/schemas/filters.py#L540-L544). That change was added in https://github.com/PrefectHQ/prefect/pull/15426, which has a lot more context on the change. I need to read up on that to understand how we need to implement the ability in the Terraform provider.

The relevant docs (https://docs.prefect.io/v3/deploy/index#concurrency-limiting) still instruct you to set `concurrency_limit` directly, and the API docs don't say it's deprecated.

I'll read up more and reach out to the backend folks if anything needs clarification.

internal/provider/resources/deployment_test.go
```diff
@@ -22,6 +22,8 @@ type deploymentConfig struct {
 	DeploymentName         string
 	DeploymentResourceName string
 
+	ConcurrencyLimit       int
```
parkedwards, line 25: should we just make this an int64, since you have to cast it

internal/api/deployments.go
```diff
@@ -75,3 +82,22 @@ type DeploymentUpdate struct {
...
 }
+
+// ConcurrencyOptions is a representation of the deployment concurrency options.
+type ConcurrencyOptions struct {
+	CollisionStrategy string `json:"collision_strategy"`
+}
+
+// GlobalConcurrencyLimit is a representation of the deployment global concurrency limit.
+type GlobalConcurrencyLimit struct {
+	Limit int `json:"limit"`
```
parkedwards, line 93: should we just make this an int64, since you're casting it

mitchnielsen: good catch, updated in 599e7e0

parkedwards (approved):
🔥

PR #307, opened 2024-11-21
Deduplicate client HTTP requests

mitchnielsen:
Add helper functions for common HTTP requests

The HTTP requests across all of our PrefectClient-compatible clients are
very similar, leading to a lot of duplicate code. This leaves room for
error when adding new clients, and makes things a bit tougher to
maintain.

This change adds helper functions to replace this redundant code, coming
with all the usual benefits of DRY-ing things up.

This is also relevant as we aim for parity with the Prefect API, and the
less code we have to write, the better!

Related to https://linear.app/prefect/issue/PLA-690/cycle-6-catch-all


internal/client/util.go
```diff
@@ -58,3 +63,106 @@ func setDefaultHeaders(request *http.Request, apiKey string) {
...
+// requestWithDecodeResponse performs an HTTP request with the provided configuration,
+// and decodes the response body into the target object.
+// It returns an error if the request fails or the decoding fails.
+func requestWithDecodeResponse(ctx context.Context, client *http.Client, cfg requestConfig, target any) error {
+	resp, err := request(ctx, client, cfg)
+	if err != nil {
+		return err
+	}
+
+	defer resp.Body.Close()
```
parkedwards, line 161: guess the only difference is that this `resp.Body.Close()` happens _**slightly**_ sooner, since it gets deferred from within the helper context + not the actual API method.

parkedwards (approved):
so so clean

internal/client/util.go
mitchnielsen: @parkedwards  That's true - I considered making the defer func a return value so the caller can call it when needed, but I started with the simpler approach and there doesn't seem to be any noticeable impact with the acceptance tests. Think it's worth doing though? I'm happy either way

parkedwards: im good to roll with it! i just remember going down this path a while ago, but the `defer` part was throwing me so i didn't end up touching the logic.  but it might not make that big of a difference

PR #305, opened 2024-11-21
feat(automations): resource support

parkedwards:
resolves https://linear.app/prefect/issue/PLA-576/tfp-automations-scaffold-resource-and-add-acceptance-tests

[Documentation on Automations](https://docs.prefect.io/v3/automate/events/automations-triggers)

This PR adds `prefect_automation` support at the resource level

Automations are polymorphic, at two different levels - the `trigger` and the `actions`.  The general structure of the schema is:

```
resource "prefect_automation" "test" {
  name = "foo"
  description = "baz"
  enabled = true

  trigger = {
    // there are 4 different trigger types: event, metric, compound, and sequence
  }

  actions = [
    { type: "run-deployment", // extra attributes here that pertain to this action type },
    { type: "pause-work-queue" },
  ]
}
```

The resource logic got pretty length, so I've split them up into the following files:
- `automation_model.go` --> this is the `tfsdk`-based struct representation of the TF config, in golang
- `automation_schema.go` --> this defines the HCL configuration schema
- `automation_resource.go` --> this has the lifecycle hooks, as well as some lengthy helpers to parse and map the automation, trigger, and action objects both to and from the TF layer

# Testing
Build the provider locally at this branch

```sh
$ pwd
terraform-provider-prefect

➜ cat ~/.terraformrc
provider_installation {

  dev_overrides {
      "prefecthq/prefect" = "/Users/edwardpark/CODE/prefect/terraform-provider-prefect/build"
  }

  # For all other providers, install them directly from their origin provider
  # registries as normal. If you omit this, Terraform will _only_ use
  # the dev_overrides block, and so no other providers will be available.
  direct {}
}
```

Navigate to the examples file

```
$ cd ./examples/resources/prefect_automation
```

Add your provider configuration

```hcl
# in resource.tf

terraform {
  required_providers {
    prefect = {
      source = "prefecthq/prefect"
    }
  }
}

provider "prefect" {
  account_id   = "<your account id>"
  workspace_id = "<your workspace id>"
}
```

```sh
$ terraform init

➜ terraform apply --auto-approve
╷
│ Warning: Provider development overrides are in effect
│
│ The following provider development overrides are set in the CLI configuration:
│  - prefecthq/prefect in /Users/edwardpark/CODE/prefect/terraform-provider-prefect/build
│
│ The behavior may therefore not match any released version of the provider and applying changes may cause the state to become incompatible with
│ published releases.
╵

[...]

PR #310, opened 2024-11-26
feat(automations): datasource support

parkedwards:
resolves https://linear.app/prefect/issue/PLA-577/tfp-automations-scaffold-datasource

followup to #305


internal/provider/datasources/automation_test.go
```diff
@@ -0,0 +1,376 @@
...
+	workspace_id = {{ .EphemeralWorkspaceResourceName }}.id
+}
+`
+
+	return helpers.RenderTemplate(tmpl, cfg)
+}
+
+//nolint:paralleltest // we use the resource.ParallelTest helper instead
+func TestAccDatasource_automation(t *testing.T) {
+	eventTriggerAutomationResourceName := testutils.NewRandomPrefixedString()
```
mitchnielsen, line 276: (no changes required) something I forgot to mention in the Resource PR: because we're using ephemeral workspaces, I don't believe we _need_ the resource names to be random

mitchnielsen (approved):
looking good from my side, great stuff. weird that the CI checks didn't run?

internal/provider/datasources/automation_test.go
parkedwards: hmm - i thought they had to be unique (even across test cases) so that they wouldn't overwrite the resource in the TF state.  i know this is the datasource test, but if we were in the resource test and used the same names, wouldn't they perform an Update()?

mitchnielsen: Fair, I may be writing tests slightly with different expectations - I assume each test case is updating an existing resource so I can test `Update` functionality. Sounds like it may just be a different use case?

parkedwards: ah interesting

PR #316, opened 2024-12-05
feat(deployments): add data source

### LINKED ISSUE #314: Add datasource support for Deployments
It would be great to add a datasource for deployments. This way it would be easy to "duplicate" a deployment into another in datasourcing it's attributes

mitchnielsen:
Implements the data source for Deployments.

Closes https://linear.app/prefect/issue/PLA-799/add-datasource-support-for-deployments

Closes https://github.com/PrefectHQ/terraform-provider-prefect/issues/314


internal/provider/datasources/deployment.go
```diff
@@ -0,0 +1,405 @@
...
+// Schema defines the scema for the data source.
+func (d *deploymentDataSource) Schema(_ context.Context, _ datasource.SchemaRequest, resp *datasource.SchemaResponse) {
+	resp.Schema = schema.Schema{
+		Description: `
+Get information about an existing Deployment by either:
+- deployment ID, or
+- deployment name
+The Deployment ID takes precedence over deployment name.
+`,
+		Attributes: map[string]schema.Attribute{
```
mitchnielsen, line 59: I'd love to just copy the attributes from the Deployment resource, but the fields on each attribute are a little different for data sources.

internal/provider/datasources/deployment.go
```diff
@@ -0,0 +1,405 @@
...
+// deploymentDataSource is the data source implementation.
+type deploymentDataSource struct {
+	client api.PrefectClient
+}
+
+// DeploymentDataSourceModel defines the Terraform data source model.
+type DeploymentDataSourceModel struct {
+	// The model requires the same fields, so reuse the fields defined in the resource model.
+	resources.DeploymentResourceModel
+}
```
mitchnielsen, line 36: afaik, this is the first time trying something like this - let me know if there are any drawbacks I may have missed

internal/provider/datasources/deployment.go
```diff
@@ -0,0 +1,405 @@
...
+		return
+	}
+
+	d.client = client
+}
+
+// copyDeploymentToModel leverages the function by the same name from the resources package to avoid repeating
+// the logic. Because DeploymentResourceModel embeds the resource.DeploymentResourceModel type, we can cast
+// it to the compatible type before calling the referenced function.
+func copyDeploymentToModel(ctx context.Context, deployment *api.Deployment, model *DeploymentDataSourceModel) diag.Diagnostics {
```
mitchnielsen, line 343: This accompanies the embedded struct for the data source model. It's a bit clunky to translate between the two types, but imo it beats repeating all of the logic in CopyDeploymentToModel. open to feedback though

jamiezieziula (approved):

internal/provider/datasources/deployment.go
parkedwards: honestly, im not sure there is a drawback yet. they use the same `types` package

internal/provider/datasources/deployment.go
```diff
@@ -0,0 +1,405 @@
...
+			},
+			"name": schema.StringAttribute{
+				Computed:    true,
+				Description: "Name of the deployment",
+				Optional:    true,
+			},
+			"flow_id": schema.StringAttribute{
+				CustomType:  customtypes.UUIDType{},
+				Description: "Flow ID (UUID) to associate deployment to",
+				Optional:    true,
```
parkedwards, line 93: aside from the user-provided inputs (name, id, account_id, workspace_id), should any attribute be `Optional` and not just read-only `Computed: true`?

internal/provider/datasources/deployment.go
```diff
@@ -0,0 +1,405 @@
...
+		return
+	}
+
+	client, err := d.client.Deployments(model.AccountID.ValueUUID(), model.WorkspaceID.ValueUUID())
+	if err != nil {
+		resp.Diagnostics.Append(helpers.CreateClientErrorDiagnostic("Deployment", err))
+	}
+
+	// A deployment can be imported + read by either ID or Handle
+	// If both are set, we prefer the ID
```
parkedwards, line 254: 👍

internal/provider/datasources/deployment.go
parkedwards: well this is pretty smart

parkedwards (approved):
reusing the resource components is smart. just had a question about Optional values in the schema - im not 100% sure we need them, but down to be wrong

internal/provider/datasources/deployment.go
mitchnielsen: was thinking we could make a new `common` package next to `resources` and `datasources` and put common elements there 🤔

internal/provider/datasources/deployment.go
mitchnielsen: good call - I understand this better now, updated in 1b35a3d (but let me know if that doesn't match what you were thinking)

parkedwards (approved):

Issue #317, opened 2024-12-06
Errors when trying to deploy an automation


qchuchu:
I'm currently receiving an issue when trying to build my automations through the terraform prefect provider (2.9.0). See my screenshot below.

### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request.
* Please do not leave _+1_ or _me too_ comments, they generate extra noise for issue followers and do not help prioritize the request.
* If you are interested in working on this issue or have submitted a pull request, please leave a comment.

### Terraform Version

Terraform version 1.5.6

### Affected Resource(s)

* prefect_automations

### Terraform Configuration Files

```tf

locals {
  alerting_slack_webhook_url = "https://hooks.slack.com/services/<redacted>"
}

resource "prefect_block" "slack_notification" {
  for_each = {
    success = "success"
    failure = "failure"
  }

  name = "dev-slack-${each.key}-notification"
  type_slug = "slack-webhook"
  data = jsonencode({
    url = local.alerting_slack_webhook_url
    notify_type = each.value
  })
}

locals {
  notification_types = {
    failure = {
      name = "[Dev] Flow Run Failure"
      description = "Event that triggers when a flow run fails"
      expect = ["prefect.flow-run.Failed"]
      block_id = prefect_block.slack_notification["failure"].id
    }
    success = {
      name = "[Dev] Flow Run Success" 
      description = "Event that triggers when a flow run succeeds"
      expect = ["prefect.flow-run.Completed"]
      block_id = prefect_block.slack_notification["success"].id
    }
  }
}

resource "prefect_automation" "flow_notifications" {
  for_each = local.notification_types

  name = each.value.name
  description = each.value.description

  trigger = {
    event = {
      posture = "Reactive"
      match = jsonencode({
        "prefect.resource.id" = "prefect.flow-run.*"
      })
      match_related = jsonencode({
        "prefect.resource.name" = [
          "prefect.deployment.dev-*"
        ],
        "prefect.resource.role" = "deployment"
      })
      after = []
      expect = each.value.expect
      for_each = [
        "prefect.resource.id"
      ]
      threshold = 1
      within = 0
    }
  }

  actions = [
    {
      block_document_id = each.value.block_id
      type = "send-notification"
      subject = "A Prefect automation '{{ automation.name }}' was triggered"
[...]

PR #318, opened 2024-12-06
fix(automations): use req.Plan in Create() for default value

parkedwards:
resolves https://linear.app/prefect/issue/PLA-803/errors-when-trying-to-deploy-an-automation

while investigating https://github.com/PrefectHQ/terraform-provider-prefect/issues/317, I noticed that the default value we set in the `prefect_automation.enabled` resource schema attribute wasn't being respected.  We set it to be `true` if no user configuration is provided, but the user submitting #317 ran into an issue where omitting the `enabled` flag was causing a state value conflict - the API was returning `false`, because we were sending a `false`

It looks like we mistakenly used `req.Config` in the `Create()` method to pull the HCL configuration, when we should be using `req.Plan`
- `req.Config` - raw, user-inputted configuration (so the `enabled` flag wouldn't be set).  Will be missing any computed/interpolated values as well
- `req.Plan` - the full computed plan, which incorporates any schema defaults

We also do this in `deployment` and `flow` resource schemas - when we look at the TF plugin framework docs, it looks like:
- [Create()](https://developer.hashicorp.com/terraform/plugin/framework/resources/create#define-create-method) uses `req.Plan`
- [Update()](https://developer.hashicorp.com/terraform/plugin/framework/resources/update#define-update-method) uses `req.Plan`
- [Read()](https://developer.hashicorp.com/terraform/plugin/framework/resources/read#define-read-method) uses `req.State`
- [Delete()](https://developer.hashicorp.com/terraform/plugin/framework/resources/delete) uses `req.State`

[Additionally, this section in the framework docs calls out](https://developer.hashicorp.com/terraform/plugin/framework/handling-data/accessing-values) using either `req.Plan` or `req.State` to access values for different situations. `req.Config` is not mentioned  

Net net, we should only use `req.Config` for certain situations when we need access to the raw user input.  Otherwise, we should be using `req.Plan` and `req.State` to reference the fully computed plan (or serialized state)

### DIFF
--- a/internal/client/util.go
+++ b/internal/client/util.go
@@ -133,7 +133,9 @@ func request(ctx context.Context, client *http.Client, cfg requestConfig) (*http
 	}
 
 	if !success {
-		return nil, fmt.Errorf("status code %s", resp.Status)
+		body, _ := io.ReadAll(resp.Body)
+
+		return nil, fmt.Errorf("status code=%s, error=%s", resp.Status, body)
 	}
 
 	return resp, nil
--- a/internal/provider/resources/deployment.go
+++ b/internal/provider/resources/deployment.go
@@ -527,7 +527,7 @@ func (r *DeploymentResource) Create(ctx context.Context, req resource.CreateRequ
 	var plan DeploymentResourceModel
 
 	// Populate the model from resource configuration and emit diagnostics on error
-	resp.Diagnostics.Append(req.Config.Get(ctx, &plan)...)
+	resp.Diagnostics.Append(req.Plan.Get(ctx, &plan)...)
 	if resp.Diagnostics.HasError() {
 		return
 	}
--- a/internal/provider/resources/deployment.go
+++ b/internal/provider/resources/deployment.go
@@ -19,6 +19,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema/listdefault"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema/listplanmodifier"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema/planmodifier"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/stringdefault"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema/stringplanmodifier"
 	"github.com/hashicorp/terraform-plugin-framework/schema/validator"
 	"github.com/hashicorp/terraform-plugin-framework/types"
@@ -238,6 +239,7 @@ func (r *DeploymentResource) Schema(_ context.Context, _ resource.SchemaRequest,
 				Optional:    true,
 				Computed:    true,
 				CustomType:  jsontypes.NormalizedType{},
+				Default:     stringdefault.StaticString("{}"),
 			},
 			"work_queue_name": schema.StringAttribute{
 				Description: "The work queue for the deployment. If no work queue is set, work will not be scheduled.",
@@ -299,12 +301,14 @@ func (r *DeploymentResource) Schema(_ context.Context, _ resource.SchemaRequest,
 				Optional:    true,
 				Computed:    true,
 				CustomType:  jsontypes.NormalizedType{},
+				Default:     stringdefault.StaticString("{}"),
 			},
 			"parameter_openapi_schema": schema.StringAttribute{
 				Description: "The parameter schema of the flow, including defaults.",
 				Optional:    true,
 				Computed:    true,
 				CustomType:  jsontypes.NormalizedType{},
+				Default:     stringdefault.StaticString("{}"),
 				// OpenAPI schema is also only set on create, and
 				// we do not support modifying this value. Therefore, any changes
 				// to this attribute will force a replacement.
@@ -315,7 +319,6 @@ func (r *DeploymentResource) Schema(_ context.Context, _ resource.SchemaRequest,
 			"concurrency_limit": schema.Int64Attribute{
 				Description: "The deployment's concurrency limit.",
 				Optional:    true,
-				Computed:    true,
 				Validators: []validator.Int64{
 					int64validator.AtLeast(1),
 				},
@@ -323,7 +326,6 @@ func (r *DeploymentResource) Schema(_ context.Context, _ resource.SchemaRequest,
 			"concurrency_options": schema.SingleNestedAttribute{
 				Description: "Concurrency options for the deployment.",
 				Optional:    true,
-				Computed:    true,
 				Attributes: map[string]schema.Attribute{
 					"collision_strategy": schema.StringAttribute{
 						Description: "Enumeration of concurrency collision strategies.",
@@ -762,7 +764,7 @@ func (r *DeploymentResource) Update(ctx context.Context, req resource.UpdateRequ
 		WorkQueueName:          model.WorkQueueName.ValueString(),
 	}
 
-	if !model.ConcurrencyOptions.CollisionStrategy.IsNull() {
+	if model.ConcurrencyOptions != nil {
 		payload.ConcurrencyOptions = &api.ConcurrencyOptions{
 			CollisionStrategy: model.ConcurrencyOptions.CollisionStrategy.ValueString(),
 		}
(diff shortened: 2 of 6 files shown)


parkedwards:
@mitchnielsen @jimid27 ok so fixing the `req.Plan` thing uncovered a couple of _**other**_ bugs in the deployment schema that i think are worth calling out here (will do it in each code line)

internal/client/util.go
```diff
@@ -133,7 +133,9 @@ func request(ctx context.Context, client *http.Client, cfg requestConfig) (*http
 	}
 
 	if !success {
-		return nil, fmt.Errorf("status code %s", resp.Status)
+		body, _ := io.ReadAll(resp.Body)
+
+		return nil, fmt.Errorf("status code=%s, error=%s", resp.Status, body)
```
parkedwards, line 138: resolves https://linear.app/prefect/issue/PLA-559/http-error-message-sometimes-not-propagated

parkedwards:
```
│ Could not create deployment, unexpected error: failed to create deployment: status code=422 Unprocessable Entity, error={"exception_message":"Invalid request
│ received.","exception_detail":[{"type":"greater_than","loc":["body","concurrency_limit"],"msg":"Input should be greater than
│ 0","input":0,"ctx":{"gt":0}}],"request_body":{"concurrency_limit":0,"flow_id":"9b824122-bb76-43c3-aded-4a629b11b76c","name":"my-deployment"}}
```

internal/provider/resources/deployment.go
```diff
@@ -238,6 +239,7 @@ func (r *DeploymentResource) Schema(_ context.Context, _ resource.SchemaRequest,
 				Optional:    true,
 				Computed:    true,
 				CustomType:  jsontypes.NormalizedType{},
+				Default:     stringdefault.StaticString("{}"),
```
parkedwards, line 242: so with `job_variable`, `parameter`, and `paramater_openai_schema` -- these are `jsontypes.NormalizedType{}`

but for the deployment_access tests, where we have HCL like this:

```
resource "prefect_deployment" "test" {
  name = "my-deployment"
  flow_id = prefect_flow.test.id
}
```

i was getting errors like this:

```
│ Error: Normalized JSON Unmarshal Error
│
│   with prefect_deployment.test,
│   on prefect_deployment.tf line 47, in resource "prefect_deployment" "test":
│   47: resource "prefect_deployment" "test" {
│
│ json string value is unknown
```

i _**think**_ what happened here was - when we swapped from req.Config -> req.Plan, we're now using the computed HCL configuration in the golang code.  So before, we're using req.Config, I want to say that it just treated unset values as `null`.  But when you actually use the resource schema via req.Plan, they become `unknown` if unset - the expectation is that the API may set something here

so when the error says "json string value is unknown", it's attempting to set what is now an unknown value from the computed plan -> golang struct that expects a json.NormalizedType

Because we want to make these an optional field, the fix here is to set a default value

internal/provider/resources/deployment.go
```diff
@@ -315,15 +319,13 @@ func (r *DeploymentResource) Schema(_ context.Context, _ resource.SchemaRequest,
 			"concurrency_limit": schema.Int64Attribute{
 				Description: "The deployment's concurrency limit.",
 				Optional:    true,
-				Computed:    true,
```
parkedwards, line 318: here's another tricky quirk.  So `concurrency_limit` is an optional value for us in the API - but when its set, it's expected to be > 0

When we used to have this as `Computed` and later switched to using req.Plan - same thing, a missing `concurrency_limit` attribute in the HCL results in an `unknown` value

Later, we attach this to the API request like so:

```golang
	createPayload := api.DeploymentCreate{
		ConcurrencyLimit:       plan.ConcurrencyLimit.ValueInt64Pointer(),
        }
```

`ValueInt64Pointer` has this documentation:

> func (i basetypes.Int64Value) ValueInt64Pointer() *int64
ValueInt64Pointer returns a pointer to the known int64 value, nil for a null value, or a pointer to 0 for an unknown value.

Because it was `unknown`, the API payload received a pointer to the value `0` - and that was being sent to the API

instead, we want to have this be `null` when its not set, so that the API call omits the key entirely

internal/provider/resources/deployment.go
```diff
@@ -762,7 +764,7 @@ func (r *DeploymentResource) Update(ctx context.Context, req resource.UpdateRequ
 		WorkQueueName:          model.WorkQueueName.ValueString(),
 	}
 
-	if !model.ConcurrencyOptions.CollisionStrategy.IsNull() {
+	if model.ConcurrencyOptions != nil {
```
parkedwards, line 767: this wasn't causing problems, but in `Create()` I noticed that we do a nil check on the parent `ConcurrencyOptions` pointer struct, rather than the nested `CollisionStrategy` attribute (since that won't scale if we need to add more attributes).  So i updated this here to match

internal/provider/resources/deployment.go
mitchnielsen: Thanks, this helps me better understand when to use Computed in general

internal/provider/resources/deployment.go
mitchnielsen: good catch, I missed that

internal/provider/resources/deployment.go
```diff
@@ -315,15 +319,13 @@ func (r *DeploymentResource) Schema(_ context.Context, _ resource.SchemaRequest,
...
 				Optional:    true,
-				Computed:    true,
 				Validators: []validator.Int64{
 					int64validator.AtLeast(1),
 				},
 			},
 			"concurrency_options": schema.SingleNestedAttribute{
 				Description: "Concurrency options for the deployment.",
 				Optional:    true,
-				Computed:    true,
```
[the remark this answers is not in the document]
mitchnielsen: Whew, weird. I wonder if we need to be making more use of `.IsUnknown()` in the resource logic (a very unfinished thought)

internal/provider/resources/deployment.go
mitchnielsen: Similar to comment below, I wonder if we need to be making better use of `IsNull()` here - because I would have avoided setting a default because I assumed it'd try to always send that field in the payload. But makes sense that if the default value looks "empty" to the `json` package, it should `omitempty` it.

internal/provider/resources/deployment.go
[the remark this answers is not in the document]
parkedwards: one more thing to add - two ways to avoid `unknown` values during plan time:

1. Optional=true, Computed=true + add a Default value

> // Default defines a proposed new state (plan) value for the attribute
	// if the configuration value is null. Default prevents the framework
	// from automatically marking the value as unknown during planning when
	// other proposed new state changes are detected. If the attribute is
	// computed and the value could be altered by other changes then a default
	// should be avoided and a plan modifier should be used instead.

OR

2. Optional=true (only) -- this makes an omitted value to be `null`

internal/client/util.go
mitchnielsen: 🤦🏼 oops. great find dude, really glad to have the full error message back

mitchnielsen (approved):

internal/provider/resources/deployment.go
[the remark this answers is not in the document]
parkedwards: > I wonder if we need to be making more use of .IsUnknown() in the resource logic (a very unfinished thought)

i think we definitely should, same with `.IsNull()`.  That said, my gut is that 90% of the time we won't need to, because null values should be treated as null API attributes in the network call

for IsUnknown, im thinking we'll use it more after this PR, since my hope is that we have a slightly better understanding of it

internal/provider/resources/deployment.go
parkedwards: >But makes sense that if the default value looks "empty" to the json package, it should omitempty it.

or maybe what this Default does is that it always sends the empty JSON object?

mitchnielsen: 🤔 I'll have to read up on this a lot more

PR #319, opened 2024-12-10
feat(deployments): support multiple schedules

mitchnielsen:
## Summary

Supports multiple schedules per deployment.

The list schedules returned by Create should only include the schedule
associated with that resource, rather than all schedules on the
deployment, so we can use that to create the schedule and then persist
the correct one to the state.

Then for Read calls, we can get the correct schedule by ID.

Related to https://linear.app/prefect/issue/PLA-772/deployments-support-multiple-schedules

## Testing

You can deploy the example configuration and check the Deployment in the UI:

image

I also tested changing one of the schedule values, `active`, and confirmed it updated the correct schedule:

image

### DIFF
--- a/internal/provider/helpers/base_model.go
+++ b/internal/provider/helpers/base_model.go
@@ -1,0 +1,11 @@
+package helpers
+
+import "github.com/prefecthq/terraform-provider-prefect/internal/provider/customtypes"
+
+// BaseModel is embedded in all other types and defines fields
+// common to all Prefect data models.
+type BaseModel struct {
+	ID      customtypes.UUIDValue      `tfsdk:"id"`
+	Created customtypes.TimestampValue `tfsdk:"created"`
+	Updated customtypes.TimestampValue `tfsdk:"updated"`
+}
--- a/internal/provider/resources/deployment_schedule.go
+++ b/internal/provider/resources/deployment_schedule.go
@@ -22,9 +22,7 @@ type DeploymentScheduleResource struct {
 }
 
 type DeploymentScheduleResourceModel struct {
-	ID      customtypes.UUIDValue      `tfsdk:"id"`
-	Created customtypes.TimestampValue `tfsdk:"created"`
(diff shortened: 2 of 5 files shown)


internal/provider/helpers/base_model.go
```diff
@@ -0,0 +1,11 @@
...
+
+import "github.com/prefecthq/terraform-provider-prefect/internal/provider/customtypes"
+
+// BaseModel is embedded in all other types and defines fields
+// common to all Prefect data models.
+type BaseModel struct {
+	ID      customtypes.UUIDValue      `tfsdk:"id"`
+	Created customtypes.TimestampValue `tfsdk:"created"`
+	Updated customtypes.TimestampValue `tfsdk:"updated"`
+}
```
mitchnielsen, line 11: This change isn't required for this PR, but saw an opportunity to not repeat these fields. If we like the approach I can come back around in a separate PR and make use of it on the rest of the {re,data}sources.

parkedwards (approved):

internal/provider/resources/deployment_schedule.go
```diff
@@ -225,6 +222,12 @@ func (r *DeploymentScheduleResource) Create(ctx context.Context, req resource.Cr
 		return
 	}
 
+	// The list of schedules returned by Create should only include the created schedule,
+	// not a list of all schedules on the deployment, so we can assume the first and only
+	// schedule in the list is the one we created.
+	//
+	// Additionally, we couldn't use getResourceByID here because of a race condition:
+	// we'd need an ID in the state to compare against, which doesn't exist yet.
```
mitchnielsen, line 230: https://github.com/PrefectHQ/prefect/blob/93d59f35c36afa3281893fd648c08623f35a80a3/src/prefect/server/models/deployments.py#L835-L865

I realized the returned schedules here should only be the ones in the request, not all of the schedules currently on the deployment. This made it possible to support multiple schedules in the provider.

PR #326, opened 2024-12-13
chore: consistently validate workspace_id

mitchnielsen:
Consistently checks if `workspace_id` is configured for endpoints that
use that field.

Adds a helper function to consistently provide this check.

Closes https://linear.app/prefect/issue/PLA-405/consistently-handle-resources-that-require-a-workspace-id

### DIFF
--- a/_about/CONTRIBUTING.md
+++ b/_about/CONTRIBUTING.md
@@ -167,3 +167,18 @@ The `tfplugindocs` CLI will:
 3. Crawl and extract all named examples in `examples/**` + add those HCL configurations into the examples section of each `.md`
 
 **NOTE:** If any documentation input files inside `examples/**` are modified, Github Actions will automatically run `make docs` and push any udpates to the working branch
+
+## Development considerations
+
+Here are some considerations to keep in mind when developing for the provider.
+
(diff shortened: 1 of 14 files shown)


mitchnielsen:
I started off with a check like this to see which endpoints require the workspace ID:

```
$ curl -s https://api.prefect.cloud/api/openapi.json | jq '.paths | keys' | grep workspace_id | head -n 5
  "/api/accounts/{account_id}/workspaces/{workspace_id}",
  "/api/accounts/{account_id}/workspaces/{workspace_id}/access/",
  "/api/accounts/{account_id}/workspaces/{workspace_id}/artifacts/",
  "/api/accounts/{account_id}/workspaces/{workspace_id}/artifacts/count",
  "/api/accounts/{account_id}/workspaces/{workspace_id}/artifacts/filter",
```

But to keep things simple, I just added/replaced that validation function in any clients that referenced a `workspace_id` in the first place - figuring that they wouldn't have ever worked if they weren't checking for a `workspace_id` already. But that API call may end up being helpful in the future so wrote it down here.

parkedwards (approved):
dope

PR #327, opened 2024-12-14
fix(service-accounts): allow for rotating forever keys

### LINKED ISSUE #325: Service Account Rotation
Can I get some clarity around API key rotation for service accounts. Is the rotate API only being called when the value of old_key_expires_in_seconds changes, not every time the plan is applied right?  So if I were to want to force a rotation I would change to 301, 299, etc?  

It would be great to have something like [keepers](https://registry.terraform.io/providers/hashicorp/random/latest/docs/resources/password#keepers-1) to make rotation less dependent on an actual functional aspect like the expiry time.

```
resource "prefect_service_account" "test1" {
  name = "test1"
  account_role_name = "Member"
  old_key_expires_in_seconds = 300
}
```

parkedwards:
resolves https://linear.app/prefect/issue/PLA-834/service-account-rotation
resolves https://github.com/PrefectHQ/terraform-provider-prefect/issues/325

## Testing

Before
```hcl
resource "prefect_service_account" "example_old_key_expires_later" {
  name                       = "my-service-account"
  api_key_keepers = {
    foo = "value-1"
    bar = "value-2"
  }
}


output "sa" {
  value = {
    id = prefect_service_account.example_old_key_expires_later.id
    key = prefect_service_account.example_old_key_expires_later.api_key
    expiry = prefect_service_account.example_old_key_expires_later.api_key_expiration
  }

  sensitive = true
}
```

```
➜ terraform output --json

{
  "sa": {
    "sensitive": true,
    "type": [
      "object",
      {
        "expiry": "string",
        "id": "string",
        "key": "string"
      }
    ],
    "value": {
      "expiry": null,
      "id": "<id-1>",
      "key": "<key-1>"
    }
  }
}
```

After
```hcl
resource "prefect_service_account" "example_old_key_expires_later" {
  name                       = "my-service-account"
  api_key_keepers = {
    dawg = "cat"
  }
}
```

```
➜ terraform output --json

{
  "sa": {
    "sensitive": true,
    "type": [
      "object",
      {
        "expiry": "string",
        "id": "string",
        "key": "string"
      }
    ],
    "value": {
      "expiry": null,
      "id": "<id-1>", <------ same ID, so SA wasn't re-created
      "key": "<key-2>" <------ different key, so SA was re-created
    }
  }
}
```

### DIFF
--- a/create-dev-testfile.sh
+++ b/create-dev-testfile.sh
@@ -31,7 +31,11 @@ terraform {
   }
 }
 
-provider "prefect" {}
+provider "prefect" {
+$([ -n "$PREFECT_API_URL" ] && echo '  endpoint = "'$PREFECT_API_URL'"')
+$([ -n "$PREFECT_API_KEY" ] && echo '  api_key = "'$PREFECT_API_KEY'"')
+$([ -n "$PREFECT_CLOUD_ACCOUNT_ID" ] && echo '  account_id = "'$PREFECT_CLOUD_ACCOUNT_ID'"')
+}
 
 resource "${resource}" "${name}" {}
 EOF
--- a/docs/resources/service_account.md
+++ b/docs/resources/service_account.md
@@ -32,9 +32,42 @@ resource "time_rotating" "ninety_days" {
 }
 # Pass the time_rotating resource to the `api_key_expiration` attribute
 # in order to automate the rotation of the Service Account key
-resource "prefect_service_account" "example" {
+resource "prefect_service_account" "example_rotate_time_key" {
   name               = "my-service-account"
   api_key_expiration = time_rotating.ninety_days.rotation_rfc3339
+}
+
+# Optionally, rotate non-expiring Service Account keys
+# using the `api_key_keepers` attribute, which is an
+# arbitrary map of values that, if changed, will
+# trigger a key rotation (but not a re-creation of the Service Account)
+resource "prefect_service_account" "example_rotate_forever_key" {
+  name               = "my-service-account"
+  api_key_expiration = null # never expires
+  api_key_keepers = {
+    foo = "value-1"
+    bar = "value-2"
+  }
+}
+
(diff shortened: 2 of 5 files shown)


create-dev-testfile.sh
```diff
@@ -31,7 +31,11 @@ terraform {
   }
 }
 
-provider "prefect" {}
+provider "prefect" {
+$([ -n "$PREFECT_API_URL" ] && echo '  endpoint = "'$PREFECT_API_URL'"')
+$([ -n "$PREFECT_API_KEY" ] && echo '  api_key = "'$PREFECT_API_KEY'"')
+$([ -n "$PREFECT_CLOUD_ACCOUNT_ID" ] && echo '  account_id = "'$PREFECT_CLOUD_ACCOUNT_ID'"')
+}
```
parkedwards, line 38: so if these env vars are set, then running `make dev-new ...` will auto-populate the generated provider block (i got tired of having to copy paste every time)

jimid27 (approved):
Lgtm

create-dev-testfile.sh
mitchnielsen: @parkedwards - as an alternative, I recently added [a script](https://github.com/PrefectHQ/terraform-provider-prefect/blob/main/scripts/testacc-dev) to facilitate this by pulling the values from 1Password in case that fits into your workflow (I ran into the same thing)

mitchnielsen: Mm, I think your approach helps when we're _not_ testing against the `github-ci-tests` workspace. I'd need to update my script to let `vault_entry` be changed to another entry.

parkedwards: right, this was just for the `make dev-new` script that generates a local (gitignored) TF dir for some manual testing.  totally an optional facility, the acceptance testing suite should be the source of truth

PR #340, opened 2024-12-18
chore(JSON attributes): rely on Normalized for object equality

mitchnielsen:
## Summary

The [JSON Normalized type mentions](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework-jsontypes/jsontypes#Normalized):

> Semantic equality logic is defined for Normalized such that
> inconsequential differences between JSON strings are ignored
> (whitespace, property order, etc).

Because of this, we should be able to rely on the normalization we do with `NewNormalizedValue` to account for any differences between the user-provided data and the data returned by the API.

This allows us to update the tests to actually validate the value stored in the resource in state, rather than only checking the API object fields.

## References
- https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework-jsontypes@v0.2.0/jsontypes


internal/provider/helpers/marshal.go
```diff
@@ -5,9 +5,10 @@ import (
 	"github.com/hashicorp/terraform-plugin-framework/diag"
 )
 
-// SafeUnmarshal is a helper function for safely unmarshalling a JSON object by checking if
-// it is null before attempting to unmarshal it. This should always be used for optional attributes.
-func SafeUnmarshal(attribute jsontypes.Normalized) (map[string]interface{}, diag.Diagnostics) {
+// UnmarshalOptional attempts to unmarshal an optional attribute. If it is null, it returns an empty
+// map[string]interface{} and no diagnostics. If it is not null, it attempts to unmarshal it and returns
+// any diagnostics.
+func UnmarshalOptional(attribute jsontypes.Normalized) (map[string]interface{}, diag.Diagnostics) {
```
mitchnielsen, line 11: This is a cosmetic change, but hopefully better communicates the purpose of this function (528475b).

internal/provider/resources/automation_test.go
```diff
@@ -322,8 +322,8 @@ func TestAccResource_automation(t *testing.T) {
 					resource.TestCheckResourceAttr(metricTriggerAutomationResourceNameAndPath, "name", "test-metric-automation"),
 					resource.TestCheckResourceAttr(metricTriggerAutomationResourceNameAndPath, "description", "description for test-metric-automation"),
 					resource.TestCheckResourceAttr(metricTriggerAutomationResourceNameAndPath, "enabled", "true"),
-					testutils.TestCheckJSONAttr(metricTriggerAutomationResourceNameAndPath, "trigger.metric.match", `{"prefect.resource.id":"prefect.flow-run.*"}`),
-					testutils.TestCheckJSONAttr(metricTriggerAutomationResourceNameAndPath, "trigger.metric.match_related", `{"prefect.resource.id":"prefect.flow-run.*","prefect.resource.role":"flow"}`),
+					resource.TestCheckResourceAttr(metricTriggerAutomationResourceNameAndPath, "trigger.metric.match", testutils.NormalizedValueForJSON(t, `{"prefect.resource.id":"prefect.flow-run.*"}`)),
+					resource.TestCheckResourceAttr(metricTriggerAutomationResourceNameAndPath, "trigger.metric.match_related", testutils.NormalizedValueForJSON(t, `{"prefect.resource.id":"prefect.flow-run.*","prefect.resource.role":"flow"}`)),
```
mitchnielsen, line 326: The changes to this automations test file use the `NormalizedValueForJSON` function added in this PR, which allowed me to remove the `TestCheckJSONAttr` function for consistency. I'm happy to restore that, though, if the existing setup was more clear (b5fc151).

internal/provider/resources/deployment.go
```diff
@@ -753,24 +770,6 @@ func (r *DeploymentResource) Read(ctx context.Context, req resource.ReadRequest,
...
-		resp.Diagnostics.Append(helpers.SerializeDataErrorDiagnostic("job_variables", "Deployment job variables", err))
-	}
-	model.JobVariables = jsontypes.NewNormalizedValue(string(jobVariablesByteSlice))
-
-	parameterOpenAPISchemaByteSlice, err := json.Marshal(deployment.ParameterOpenAPISchema)
-	if err != nil {
-		resp.Diagnostics.Append(helpers.SerializeDataErrorDiagnostic("parameter_openapi_schema", "Deployment parameter OpenAPI schema", err))
-	}
-	model.ParameterOpenAPISchema = jsontypes.NewNormalizedValue(string(parameterOpenAPISchemaByteSlice))
-
```
mitchnielsen, line 773: There were two identical sections like this, so I moved it to the `CopyDeploymentToModel` function (2a0396c).

internal/provider/resources/work_pool_test.go
```diff
@@ -77,7 +72,8 @@ func TestAccResource_work_pool(t *testing.T) {
 				Config: fixtureAccWorkPoolCreate(workspace.Resource, randomName, poolType, baseJobTemplate, true),
 				Check: resource.ComposeAggregateTestCheckFunc(
 					testAccCheckWorkPoolExists(workPoolResourceName, &workPool),
-					testAccCheckWorkPoolValues(&workPool, &api.WorkPool{Name: randomName, Type: poolType, BaseJobTemplate: baseJobTemplateMap, IsPaused: true}),
+					testAccCheckWorkPoolValues(&workPool, &api.WorkPool{Name: randomName, Type: poolType, IsPaused: true}),
+					resource.TestCheckResourceAttr(workPoolResourceName, "base_job_template", baseJobTemplateExpected),
```
mitchnielsen, line 76: Instead of checking the values of the deployment, we can now go one step further and also test that the value stored in the resource via state is what we expect (9ff43f3).

internal/provider/helpers/compare.go
```diff
@@ -1,14 +0,0 @@
-package helpers
-
-import "github.com/go-test/deep"
-
-// ObjectsEqual checks to see if two objects are equivalent, accounting for
-// differences in the order of the contents.
-func ObjectsEqual(obj1, obj2 interface{}) (bool, []string) {
```
mitchnielsen, line 7: This is replaced by `NormalizedValueForJSON`, where we rely on the `jsontypes.Normalized` type to calculate equality.

mitchnielsen:
Will open up for review since it's ready for that, but it's non-urgent so no rush here

jamiezieziula (approved):

parkedwards (approved):
fire

Issue #330, opened 2024-12-16
Can't Datasource a Deployment by its name


phergoualch:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request.
* Please do not leave _+1_ or _me too_ comments, they generate extra noise for issue followers and do not help prioritize the request.
* If you are interested in working on this issue or have submitted a pull request, please leave a comment.

### Terraform Version

1.9.8

### Affected Resource(s)

* data.prefect_deployment

### Terraform Configuration Files

```tf
data "data" "existing" {
  # id = "9e1eb232-4f15-404d-8a84-3727cc68f584"
  name = "mysql-backup/mysql-backup"
}
```

### Expected Behavior

It is working fine with the deployment id but not the deployment name. I tested both **mysql-backup** and **mysql-backup/mysql-backup** as outputed by the prefect deployment ls.

$ prefect deployment ls

| Name | ID | Work Pool |
|------|------|------------|
| mysql-backup/mysql-backup | 9e1eb232-4f15-404d-8a84-3727cc68f584 | ecs-fargate-spot |

mitchnielsen:
Thanks for reaching out @phergoualch

I believe you just need to tweak your `data` block slightly based on the [documentation](https://registry.terraform.io/providers/PrefectHQ/prefect/latest/docs/data-sources/deployment):

```diff
- data "data" "existing" {
+ data "prefect_deployment" "existing"
  # id = "9e1eb232-4f15-404d-8a84-3727cc68f584"
-  name = "mysql-backup/mysql-backup"
+  name = "mysql-backup"
}
```

This does the following:
- Changes the first argument to "data" from "data" to "prefect_deployment" so it knows which data source logic to use
- Changes the name to the name of the deployment (I believe `prefect deployment ls` prints out `<flow name>/<deployment name>`)

phergoualch:
Well, the data is a mistake from my copy past, I used the correct data_source name.
And as said in the issue, I tested both **mysql-backup** and **mysql-backup/mysql-backup**

mitchnielsen:
Sure, just pointing out that the documentation shows you only need the deployment name, not `flow name/deployment name`.

Since I wasn't able to replicate the error, can you please:
1. Include your complete (relevant) Terraform configuration instead of just the `data` block
[...]

PR #343, opened 2024-12-19
fix(deployments datasource): get by name

mitchnielsen:
## Summary

For the Deployment datasource, we were getting it by name by using the
/filter endpoint. But we weren't passing any filter body in the request,
so it was always getting all Deployments back.

This change uses the API endpoint to get a deployment by name, which now
requires a flow name as well.

Related to https://github.com/PrefectHQ/terraform-provider-prefect/issues/330

### DIFF
--- a/internal/client/deployments.go
+++ b/internal/client/deployments.go
@@ -60,29 +60,30 @@ func (c *DeploymentsClient) Create(ctx context.Context, data api.DeploymentCreat
 	return &deployment, nil
 }
 
-// List returns a list of Deployments based on the provided list of names.
-func (c *DeploymentsClient) List(ctx context.Context, _ []string) ([]*api.Deployment, error) {
-	cfg := requestConfig{
-		method:       http.MethodPost,
-		url:          fmt.Sprintf("%s/filter", c.routePrefix),
-		body:         http.NoBody,
-		apiKey:       c.apiKey,
-		successCodes: successCodesStatusOK,
-	}
-
-	var deployments []*api.Deployment
-	if err := requestWithDecodeResponse(ctx, c.hc, cfg, &deployments); err != nil {
-		return nil, fmt.Errorf("failed to list deployments: %w", err)
-	}
-
-	return deployments, nil
-}
-
 // Get returns details for a Deployment by ID.
 func (c *DeploymentsClient) Get(ctx context.Context, deploymentID uuid.UUID) (*api.Deployment, error) {
 	cfg := requestConfig{
 		method:       http.MethodGet,
 		url:          fmt.Sprintf("%s/%s", c.routePrefix, deploymentID.String()),
+		body:         http.NoBody,
+		apiKey:       c.apiKey,
+		successCodes: successCodesStatusOK,
+	}
+
+	var deployment api.Deployment
+	if err := requestWithDecodeResponse(ctx, c.hc, cfg, &deployment); err != nil {
+		return nil, fmt.Errorf("failed to get deployment: %w", err)
+	}
+
+	return &deployment, nil
+}
+
+// GetByName returns details for a Deployment by name.
+func (c *DeploymentsClient) GetByName(ctx context.Context, flowName, deploymentName string) (*api.Deployment, error) {
+	url := fmt.Sprintf("%s/name/%s/%s", c.routePrefix, flowName, deploymentName)
+	cfg := requestConfig{
+		method:       http.MethodGet,
+		url:          url,
 		body:         http.NoBody,
 		apiKey:       c.apiKey,
 		successCodes: successCodesStatusOK,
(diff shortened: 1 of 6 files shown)


internal/client/deployments.go
```diff
@@ -60,29 +60,30 @@ func (c *DeploymentsClient) Create(ctx context.Context, data api.DeploymentCreat
...
 
-	return deployments, nil
+	return &deployment, nil
 }
 
-// Get returns details for a Deployment by ID.
-func (c *DeploymentsClient) Get(ctx context.Context, deploymentID uuid.UUID) (*api.Deployment, error) {
+// GetByName returns details for a Deployment by name.
+func (c *DeploymentsClient) GetByName(ctx context.Context, flowName, deploymentName string) (*api.Deployment, error) {
+	url := fmt.Sprintf("%s/name/%s/%s", c.routePrefix, flowName, deploymentName)
```
mitchnielsen, line 83: It surprised me that the [API requires the flow_name to get a deployment by name](https://app.prefect.cloud/api/docs#tag/Deployments/operation/read_deployment_by_name_api_accounts__account_id__workspaces__workspace_id__deployments_name__flow_name___deployment_name__get). I started by adding that field as an optional attribute on the datasource specifically, and then validate that either id or name and flow_name are set.

Alternatives considered so far:
- Going back to `List`/`Filter`, and passing the filter payload: the [filter payload](https://app.prefect.cloud/api/docs#tag/Deployments/operation/read_deployments_api_accounts__account_id__workspaces__workspace_id__deployments_filter_post) is pretty complex, and I'm not keen to model the payload and response in Terraform tbh
- Going back to `List`/`Filter`, and getting all Deployments then finding the one with a matching name: this doesn't quite feel right given the potential size of the response, and I have this feeling that `flow_name` is required for a reason - maybe Deployments can have the same name if they're associated with different flows?
- Automatically getting the `flow_name` for the user based on the `flow_id`: we already have the `flow_id` set on the Deployment resource, so this would be easier for users in some cases (such as when they don't define the flow in Terraform and therefore don't have access to `prefect_flow.example.name`). This is a fairly simple API call, and is probably my preference out of these options so far, but still doesn't quite feel optimal.

Open to any opinions here

internal/client/deployments.go
```diff
@@ -60,29 +60,30 @@ func (c *DeploymentsClient) Create(ctx context.Context, data api.DeploymentCreat
 	return &deployment, nil
 }
 
-// List returns a list of Deployments based on the provided list of names.
-func (c *DeploymentsClient) List(ctx context.Context, _ []string) ([]*api.Deployment, error) {
+// Get returns details for a Deployment by ID.
+func (c *DeploymentsClient) Get(ctx context.Context, deploymentID uuid.UUID) (*api.Deployment, error) {
```
mitchnielsen, line 64: This diff shows things in kind of a weird way. I technically replaced `List` with `GetByName`.

So to clarify:
- ` Get` is unchanged
- `GetByName` is added
- `List` is removed

internal/client/deployments.go
parkedwards: IMO, the option we choose will depend on what interface we're choosing to expose to the user, which is coupled to what we think they're going to use the datasource for:

- search a single deployment by name => return a single Deployment.  since deployment names aren't unique + need an extra piece of info to triangulate (the flow itself), we could either implicitly look up the flow as you suggested OR expect the user to include that flow identifier in the datasource schema (so they'd put in a deployment name + flow name) and pass it into the filter call (see next bullet about the Deployment filter object)
- search all deployments with a name => return a list of Deployments. use the /filter endpoint, and model out the payload.  The Deployment filter schema is pretty big and includes a bunch of searchable options (work pool, flow name, etc.) so I'd advocate for investing in defining this.  We don't have to model the entire filter payload to start, it could look something like this

```golang
type DeploymentFilter struct {
	Deployments struct {
		Name struct {
			Any []string `json:"any_"`
		} `json:"name"`
	} `json:"deployments"`
}

# usage:
filter := DeploymentFilter{}
filter.Deployments.Name = []string{"deployment-1", "deployment-2"}
```

I don't think the "find all deployments and do the filtering at runtime" is better than either of those options, just given that we have a filtering endpoint that pretty much does this if we're able to represent the filtering payload

mitchnielsen: Great points - I could see that advanced datasource logic being really helpful. I created https://github.com/PrefectHQ/terraform-provider-prefect/issues/344 to track that. I think for now it makes sense to proceed with the current PR since it addresses the bug based on how we currently say this datasource should work. That's just my 2 cents though, happy to pivot

internal/provider/datasources/deployment_test.go
```diff
@@ -44,6 +44,7 @@ data "prefect_deployment" "test_by_id" {
 
 data "prefect_deployment" "test_by_name" {
   name = prefect_deployment.test.name
+  flow_name = prefect_flow.test.name
```
mitchnielsen, line 47: I suspect this was only passing before because there was only one deployment in the test workspace, meaning the existing logic actually worked. This obviously breaks in the real world when there is more than one deployment (of any name).

parkedwards (approved):

PR #345, opened 2024-12-20
fix(blocks): always use user-provided HCL for .data attr

### LINKED ISSUE #341: Blocks using credential blocks update on every plan
I am finding that my S3 Bucket block that references my AWS Credentials block is updating on every plan.  My code looks like this:

```terraform
resource "prefect_block" "prefect_aws" {
  name      = "prefect--aws"
  type_slug = "aws-credentials"

  data = jsonencode({
    aws_access_key_id     = aws_iam_access_key.prefect_aws.id
    aws_secret_access_key = aws_iam_access_key.prefect_aws.secret
    region_name           = data.aws_region.current.name
  })
}

resource "prefect_block" "my_prefect_results" {
  name      = "my-prefect-results"
  type_slug = "s3-bucket"

  data = jsonencode({
    bucket_name = "my-prefect-results"
    credentials = { "$ref" : { "block_document_id" : prefect_block.prefect_aws.id } }
  })
}

```

which always seems to yield this update on every plan:

```diff
OpenTofu will perform the following actions:

  # prefect_block.my_prefect_results will be updated in-place
! resource "prefect_block" "my_prefect_results" {
!       data      = (sensitive value)
        id        = "XXXXXXXXXXXXXXXXXXXXXXXXXXXXX"
        name      = "my-prefect-results"
!       updated   = "2024-12-10T11:37:39Z" -> (known after apply)
        # (2 unchanged attributes hidden)
    }

Plan: 0 to add, 1 to change, 0 to destroy.
```

This is just because of the data field right?  Or am I missing something.  It's only the s3 bucket block that this is happening for.

parkedwards:
resolves https://github.com/PrefectHQ/terraform-provider-prefect/issues/341
resolves https://linear.app/prefect/issue/PLA-860/blocks-using-credential-blocks-update-on-every-plan

normally, we reconcile the API values for `.data` in a block / block_document - this means that the remote value always be fetched (even on Reads) and be compared to the state values

however, when using a reference object (`$ref`), the HCL will always look like this:

```hcl
resource "prefect_block" "my_prefect_results" {
  name      = "my-prefect-results"
  type_slug = "s3-bucket"

  data = jsonencode({
    bucket_name = "my-prefect-results"
    credentials = { "$ref" : { "block_document_id" : prefect_block.prefect_aws.id } }
  })
}
```

but the API response will always come back hydrated:

```json
  "data": {
    "bucket_name": "my-prefect-results",
    "credentials": {
      "aws_access_key_id": "access-key-id",
      "aws_secret_access_key": "********",
      "region_name": "us-east-1"
    }
  },
```

this causes a no-op change to ALWAYS detect a plan update

```
  ~ update in-place

  # prefect_block.my_prefect_results will be updated in-place
  ~ resource "prefect_block" "my_prefect_results" {
      ~ data      = (sensitive value)
        id        = "0551754e-cdca-4a45-8926-6b4c9c1f927b"
        name      = "my-prefect-results"
      ~ updated   = "2024-12-20T04:26:28Z" -> (known after apply)
        # (2 unchanged attributes hidden)
    }

```

before this PR, we already omitted persisting the API value for `.data` in both `Create()` and `Update()` calls for similar reasons (Create will always return masked data, so it will always conflict with the user provided values.  Update runs into the `$ref` issue)

here, we'll omit persisting the API `.data` across the board.  this means that the user HCL is the source of truth.  The tradeoff here is that remote changes (eg. someone goes to the Prefect UI and updates a Block value) won't get pulled down during plans to detect drift -- management of this attribute becomes one direction only

### DIFF
--- a/docs/resources/block.md
+++ b/docs/resources/block.md
@@ -8,6 +8,7 @@ description: |-
   Note: you should be on version 3.0.0rc1 or later to use the following commands:
   Use prefect block type ls to view all available Block type slugs, which is used in the type_slug attribute.
   Use prefect block type inspect <slug> to view the data schema for a given Block type. Use this to construct the data attribute value (as JSON string).
+  NOTE: if a Block is managed in Terraform, the .data attribute will NOT be re-reconciled if the remote value is changed. This means that a TF-managed Block will only update the API, and not the other way around.
 ---
 
 # prefect_block (Resource)
@@ -17,6 +18,7 @@ The resource `block` allows creating and managing [Prefect Blocks](https://docs.
 *Note:* you should be on version `3.0.0rc1` or later to use the following commands:
 Use `prefect block type ls` to view all available Block type slugs, which is used in the `type_slug` attribute.
 Use `prefect block type inspect <slug>` to view the data schema for a given Block type. Use this to construct the `data` attribute value (as JSON string).
+*NOTE:* if a Block is managed in Terraform, the `.data` attribute will NOT be re-reconciled if the remote value is changed. This means that a TF-managed Block will only update the API, and not the other way around.
 
 ## Example Usage
 
--- a/internal/provider/resources/block.go
+++ b/internal/provider/resources/block.go
@@ -2,6 +2,7 @@ package resources
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"strings"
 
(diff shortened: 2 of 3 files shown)


jamiezieziula (approved):

mitchnielsen (approved):
good find, looks good to me. do you think we need to mention this in the docs? (specifically that the config is the source of truth now)

Issue #120, opened 2023-11-10
tests: investigate why variables stored during custom TestCheckFunc are zeros for built-in check functions / import checks


parkedwards:

mitchnielsen:
Resolved: [https://github.com/PrefectHQ/terraform-provider-prefect/pull/347#discussion_r1898041779](https://github.com/PrefectHQ/terraform-provider-prefect/pull/347#discussion_r1898041779)

parkedwards:
@mitchnielsen 🙏 thank you!

PR #347, opened 2024-12-24
feat(client): add GetEndpointHost utility getter

### LINKED ISSUE #333: extract path-suffix-less API endpoint into webhook resource
we're hardcoding it right now for the purposes of outputting an immediately useable `.endpoint`

parkedwards:
resolves https://linear.app/prefect/issue/PLA-841/extract-path-suffix-less-api-endpoint-into-webhook-resource
resolves #333 

this PR adds an endpointHost (host only, so no prefix) attribute to the PrefectClient, as well as a getter method on the interface.  we use this to construct the full webhook endpoint, since the API only stashes the slug portion, and we'll want to expose a fully formed endpoint based on the API that the provider is initialized with

### DIFF
--- a/internal/api/client.go
+++ b/internal/api/client.go
@@ -6,8 +6,10 @@ import "github.com/google/uuid"
 //
 //nolint:interfacebloat // we'll accept a larger PrefectClient interface
 type PrefectClient interface {
+	// Utility methods on the Client interface
 	GetEndpointHost() string
 
+	// API Client Factories - for instantiating a client for each API resource
 	Accounts(accountID uuid.UUID) (AccountsClient, error)
 	Automations(accountID uuid.UUID, workspaceID uuid.UUID) (AutomationsClient, error)
 	AccountMemberships(accountID uuid.UUID) (AccountMembershipsClient, error)
--- a/internal/api/client.go
+++ b/internal/api/client.go
@@ -6,6 +6,8 @@ import "github.com/google/uuid"
 //
 //nolint:interfacebloat // we'll accept a larger PrefectClient interface
 type PrefectClient interface {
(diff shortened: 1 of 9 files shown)


internal/api/client.go
```diff
@@ -6,6 +6,8 @@ import "github.com/google/uuid"
 //
 //nolint:interfacebloat // we'll accept a larger PrefectClient interface
 type PrefectClient interface {
+	GetEndpointHost() string
```
parkedwards, line 9: todo: add a comment string that differentiates the client factories from a new category of methods like this one

internal/provider/resources/webhooks_test.go
```diff
@@ -131,6 +134,23 @@ func testAccCheckWebhookExists(webhookResourceName string, webhook *api.Webhook)
 	}
 }
 
+func testAccCheckWebhookEndpoint(webhookResourceName string, webhook *api.Webhook) resource.TestCheckFunc {
```
parkedwards, line 137: it's worth calling out here, since i couldn't verify this behavior from their docs + only found it through testing the actual control flow

basically, there's a pattern in the testing docs where you can instantiate a pointer outside of the test steps, hydrate it with a custom test function, and then reference it again in the subsequent function

https://developer.hashicorp.com/terraform/plugin/testing/acceptance-tests/teststep#custom-check-functions

however, the nuance is that these custom functions seem to run well after any pre-built check function on the `resource.` object, eg. `resource.TestCheckResourceAttr` - so you can't run these resource checks against the hydrated value, because the pointer will still be a zero value

this answers the issue raised here, so closing this PR will resolve #120

mitchnielsen (approved):

PR #357, opened 2025-01-15
fix(deployments): align omitempty usage

### LINKED ISSUE #356: `manifest_path` causes deployment update to fail (Prefect Server)
This line of code did not omitempty the manifest field, causing this error

https://github.com/PrefectHQ/terraform-provider-prefect/blob/90ce565433daccc51e6ba51b7ae3cc393c332e8c/internal/api/deployments.go#L77

```bash
│ Error: Error updating deployment
│ 
│   with prefect_deployment.dep["run_pyspark_job"],
│   on main.tf line 74, in resource "prefect_deployment" "dep":
│   74: resource "prefect_deployment" "dep" {
│ 
│ Could not update deployment, unexpected error: failed to update deployment: status code=422 Unprocessable Entity, error={"exception_message":"Invalid request
│ received.","exception_detail":[{"type":"extra_forbidden","loc":["body","manifest_path"],"msg":"Extra inputs are not
│ permitted","input":""}]

```

mitchnielsen:
Aligns the omitempty usage with the expectations from the API.

Required fields should not set omitempty.
Optional fields should set omitempty, when we don't want to send empty values.

We don't need omitempty on the object struct, just for structs used in API calls that get (un)marshalled.

Closes https://linear.app/prefect/issue/PLA-919/manifest-path-causes-deployment-update-to-fail-prefect-server

Closes https://github.com/PrefectHQ/terraform-provider-prefect/issues/356

### DIFF
--- a/internal/api/deployments.go
+++ b/internal/api/deployments.go
@@ -22,25 +22,25 @@ type Deployment struct {
 	WorkspaceID uuid.UUID `json:"workspace_id"`
 
 	ConcurrencyLimit       *int64                  `json:"concurrency_limit"`
-	ConcurrencyOptions     *ConcurrencyOptions     `json:"concurrency_options,omitempty"`
-	Description            string                  `json:"description,omitempty"`
+	ConcurrencyOptions     *ConcurrencyOptions     `json:"concurrency_options"`
+	Description            string                  `json:"description"`
 	EnforceParameterSchema bool                    `json:"enforce_parameter_schema"`
 	Entrypoint             string                  `json:"entrypoint"`
 	FlowID                 uuid.UUID               `json:"flow_id"`
 	GlobalConcurrencyLimit *GlobalConcurrencyLimit `json:"global_concurrency_limit"`
-	JobVariables           map[string]interface{}  `json:"job_variables,omitempty"`
-	ManifestPath           string                  `json:"manifest_path,omitempty"`
+	JobVariables           map[string]interface{}  `json:"job_variables"`
+	ManifestPath           string                  `json:"manifest_path"`
 	Name                   string                  `json:"name"`
-	ParameterOpenAPISchema map[string]interface{}  `json:"parameter_openapi_schema,omitempty"`
-	Parameters             map[string]interface{}  `json:"parameters,omitempty"`
+	ParameterOpenAPISchema map[string]interface{}  `json:"parameter_openapi_schema"`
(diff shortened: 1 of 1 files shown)


internal/api/deployments.go
```diff
@@ -74,7 +74,7 @@ type DeploymentUpdate struct {
 	EnforceParameterSchema bool                   `json:"enforce_parameter_schema,omitempty"`
 	Entrypoint             string                 `json:"entrypoint,omitempty"`
 	JobVariables           map[string]interface{} `json:"job_variables,omitempty"`
-	ManifestPath           string                 `json:"manifest_path"`
+	ManifestPath           string                 `json:"manifest_path,omitempty"`
```
mitchnielsen, line 77: This is the fix for https://github.com/PrefectHQ/terraform-provider-prefect/issues/356 specifically.

internal/api/deployments.go
```diff
@@ -50,10 +50,10 @@ type DeploymentCreate struct {
 	Description            string                 `json:"description,omitempty"`
 	EnforceParameterSchema bool                   `json:"enforce_parameter_schema,omitempty"`
 	Entrypoint             string                 `json:"entrypoint,omitempty"`
-	FlowID                 uuid.UUID              `json:"flow_id"`
+	FlowID                 uuid.UUID              `json:"flow_id"` // required
```
mitchnielsen, line 53: There's no `required` option (that I'm aware of), so I added `// required` here to signify that we're intentionally not specifying `omitempty`. My thought process was this may help us in the future to determine whether the field was left off intentionally or not - open to opinions though

internal/api/deployments.go
```diff
@@ -22,25 +22,25 @@ type Deployment struct {
 	WorkspaceID uuid.UUID `json:"workspace_id"`
 
 	ConcurrencyLimit       *int64                  `json:"concurrency_limit"`
-	ConcurrencyOptions     *ConcurrencyOptions     `json:"concurrency_options,omitempty"`
-	Description            string                  `json:"description,omitempty"`
+	ConcurrencyOptions     *ConcurrencyOptions     `json:"concurrency_options"`
```
mitchnielsen, line 25: AFAIK, there's no need to specify `omitempty` on the Deployment object itself because it's not used in any API requests and is therefore never (un)marshalled by the `json` package.

jamiezieziula (approved):

parkedwards (approved):

Issue #280, opened 2024-10-07
Add support for concurrency limits


mitchnielsen:
## Summary

Let's add support for managing concurrency limits via the Terraform provider.

## Resources

- https://docs.prefect.io/3.0/develop/global-concurrency-limits
- https://app.prefect.cloud/api/docs#tag/Concurrency-Limits

ubiquitousbyte:
This will also be extremely helpful for us as well.

mitchnielsen:
Thanks all for raising this one, we've added this to the schedule for next week 👍🏼

mitchnielsen:
We've added support for a `prefect_concurrency_limit` resource in [2.14.0](https://github.com/PrefectHQ/terraform-provider-prefect/releases/tag/v2.14.0). Support for a matching datasource is planned 👍🏼

PR #360, opened 2025-01-28
chore/concurrency-limit-support

jimid27:
- Relates to: https://linear.app/prefect/issue/PLA-383/add-support-for-concurrency-limits
- Relates to: https://github.com/PrefectHQ/terraform-provider-prefect/issues/280
- Resolves: https://linear.app/prefect/issue/PLA-971/create-concurrency-limit-resource
- Creates resources for concurrency limits as defined in https://app.prefect.cloud/api/docs#tag/Concurrency-Limits


internal/provider/resources/concurrency_limit.go
```diff
@@ -0,0 +1,263 @@
...
+				Description: "Tag",
+				PlanModifiers: []planmodifier.String{
+					// Concurrency limit updates are not supported so any changes to the tag will
+					// require a replacement of the resource.
+					stringplanmodifier.RequiresReplace(),
+				},
+			},
+			"concurrency_limit": schema.Int64Attribute{
+				Required:    true,
+				Description: "Concurrency limit",
```
mitchnielsen, line 118:
```suggestion
				Description: "The concurrency limit.",
```

Small but may as well match the API docs field while we're here

internal/provider/resources/concurrency_limit.go
```diff
@@ -0,0 +1,263 @@
...
+				CustomType:  customtypes.UUIDType{},
+			},
+			"workspace_id": schema.StringAttribute{
+				Optional:    true,
+				Description: "Workspace ID (UUID)",
+				CustomType:  customtypes.UUIDType{},
+			},
+			"tag": schema.StringAttribute{
+				Required:    true,
+				Description: "Tag",
```
mitchnielsen, line 109:
```suggestion
				Description: "A tag the concurrency limit is applied to.",
```

Because this is the source for the documentation that's generated and pushed to the TF provider website, let's fill in the description from the API docs

internal/provider/resources/concurrency_limit.go
```diff
@@ -0,0 +1,263 @@
...
+}
+
+// ImportState imports the resource into Terraform state.
+func (r *ConcurrencyLimitResource) ImportState(ctx context.Context, req resource.ImportStateRequest, resp *resource.ImportStateResponse) {
+	// we'll allow input values in the form of:
+	// - "id,workspace_id"
+	// - "id"
+	maxInputCount := 2
+	inputParts := strings.Split(req.ID, ",")
+
```
mitchnielsen, line 241:
```suggestion

	// eg. ",foo" or "foo,"
	if len(inputParts) == maxInputCount && (inputParts[0] == "" || inputParts[1] == "") {
		resp.Diagnostics.AddError(
			"Unexpected Import Identifier",
			fmt.Sprintf("Expected non-empty import identifiers, in the form of `id,workspace_id`. Got %q", req.ID),
		)

		return
	}
```

This will catch if the user provides an empty identifier in one of the positions (we should really make this a reusable helper instead of defining it in every resource but this works for now)

internal/provider/resources/concurrency_limit_test.go
```diff
@@ -0,0 +1,57 @@
...
+func TestAccResource_concurrency_limit(t *testing.T) {
+	resourceName := "prefect_concurrency_limit.concurrency_limit"
+	workspace := testutils.NewEphemeralWorkspace()
+
+	resource.ParallelTest(t, resource.TestCase{
+		ProtoV6ProviderFactories: testutils.TestAccProtoV6ProviderFactories,
+		PreCheck:                 func() { testutils.AccTestPreCheck(t) },
+		Steps: []resource.TestStep{
+			{
+				// Check creation + existence of the flow resource
```
mitchnielsen, line 33:
```suggestion
				// Check creation + existence of the resource
```

mitchnielsen (commented):
nice work dude, couple small changes then we should be good to release 👍🏼

internal/provider/resources/concurrency_limit.go
```diff
@@ -0,0 +1,263 @@
...
+		return
+	}
+
+	r.client = client
+}
+
+// Schema defines the schema for the resource.
+func (r *ConcurrencyLimitResource) Schema(_ context.Context, _ resource.SchemaRequest, resp *resource.SchemaResponse) {
+	resp.Schema = schema.Schema{
+		Description: "The resource `concurrency_limit` represents a concurrency limit for a tag.",
```
mitchnielsen, line 74:
```suggestion
		Description: "The resource `concurrency_limit` represents a concurrency limit. Concurrency limits allow you to manage execution efficiently, controlling how many tasks, flows, or other operations can run simultaneously. They are ideal for optimizing resource usage, preventing bottlenecks, and customizing task execution.",
```

I wasn't sure "for a tag" was totally right, so I grabbed this description from https://docs.prefect.io/v3/develop/global-concurrency-limits#manage-global-concurrency-and-rate-limits - open to tweaking it

internal/client/concurrency_limits.go
```diff
@@ -0,0 +1,95 @@
...
+		routePrefix: getWorkspaceScopedURL(c.endpoint, accountID, workspaceID, "concurrency_limits"),
+		apiKey:      c.apiKey,
+	}, nil
+}
+
+// Create creates a new concurrency limit.
+func (c *ConcurrencyLimitsClient) Create(ctx context.Context, data api.ConcurrencyLimitCreate) (*api.ConcurrencyLimit, error) {
+	cfg := requestConfig{
+		method:       http.MethodPost,
+		url:          c.routePrefix,
```
mitchnielsen, line 48:
```suggestion
		url:          c.routePrefix + "/",
```

Small change here related to https://linear.app/prefect/issue/PLA-948/workspace-create-fails-with-404-when-endpoint-is-set

mitchnielsen (approved):

jbnitorum:
Is this for global concurrency limits or task run concurrency limits?  I see the API docs look vague as to which they apply to.  The former has the additional option of slot decay per second

mitchnielsen:
Thanks for reaching out, we had a look at the API docs and the correct endpoint in this case _appears_ to actually be the one for Concurrency Limits v2. We're confirming with the backend team and getting a better understanding and then will follow up with change(s) to make this clearer and better supported.

mitchnielsen:
We've opened https://github.com/PrefectHQ/terraform-provider-prefect/pull/362 to specify that this is for task run concurrency limits. Should have it released soon, and global concurrency limit support will come after that.

mitchnielsen:
That change is now available in [v2.14.1](https://github.com/PrefectHQ/terraform-provider-prefect/releases/tag/v2.14.1).

PR #362, opened 2025-01-31
fix(concurrency limits): rename and provide examples of task run concurrency limits

jimid27:
# Summary
- In PR https://github.com/PrefectHQ/terraform-provider-prefect/pull/360 we introduced `concurrency limits` without much supporting information.  What we were actually introducing was `task run concurrency limits`, which should have been explained to end users. 

# References
https://app.prefect.cloud/api/docs#tag/Concurrency-Limits

Related to https://linear.app/prefect/issue/PLA-383/add-support-for-concurrency-limits

# Notes
- The concurrency_limits API only supports creating and destroying `task run concurrency limits`.  These concurrency limits are referenced in `task` definitions and can be called by matching the tag defined in the `task run concurrency limit` to the tags defined in each `task` definition.  (See example)
- Task run concurrency limits cannot be updated in place and need to be recreated each time they are updated.

# Tests
- Added unit test to verify that the `task run concurrency limit` can be created and updated via destroy and recreate.


internal/provider/resources/task_run_concurrency_limit.go
```diff
@@ -69,14 +69,14 @@ func (r *ConcurrencyLimitResource) Configure(_ context.Context, req resource.Con
 }
 
 // Schema defines the schema for the resource.
-func (r *ConcurrencyLimitResource) Schema(_ context.Context, _ resource.SchemaRequest, resp *resource.SchemaResponse) {
+func (r *TaskRunConcurrencyLimitResource) Schema(_ context.Context, _ resource.SchemaRequest, resp *resource.SchemaResponse) {
 	resp.Schema = schema.Schema{
-		Description: "The resource `concurrency_limit` represents a concurrency limit. Concurrency limits allow you to manage execution efficiently, controlling how many tasks, flows, or other operations can run simultaneously. They are ideal for optimizing resource usage, preventing bottlenecks, and customizing task execution.",
+		Description: "The resource `task_run_concurrency_limit` represents a task run concurrency limit. Task Run concurrency limits  allow you to control how many tasks with specific tags can run simultaneously.",
```
mitchnielsen, line 74:
```suggestion
		Description: "The resource `task_run_concurrency_limit` represents a task run concurrency limit. Task run concurrency limits allow you to control how many tasks with specific tags can run simultaneously. For more information, see https://docs.prefect.io/v3/develop/task-run-limits.",
```

Let's link right to the docs here, what do you think?

jimid27: Yeah good call

mitchnielsen (approved):
Looks good, thanks for the quick fix on this one

PR #363, opened 2025-01-31
feat(global concurrency limit): resource support

jimid27:
# Summary
Introduces the `global_concurrency_limit` terraform resource to the terraform provider allowing users to create, update, read, and delete `global concurrency limits` from their workspaces.

More information about global concurrency limits and its uses can be found [here](https://docs.prefect.io/v3/develop/global-concurrency-limits#apply-global-concurrency-and-rate-limits)

# References
https://app.prefect.cloud/api/docs#tag/Concurrency-Limits-V2
https://docs.prefect.io/v3/develop/global-concurrency-limits#apply-global-concurrency-and-rate-limits
Resolves: https://linear.app/prefect/issue/PLA-979/create-global-concurrency-limit-resource

# Notes
 - This PR specifically adds the `global_concurrency_limit` `resource` with the `datasource` coming in a future PR during our next sprint cycle.
 - A required change to the `deployment` resource was required as one of the returned values to the deployment api upon creation is the `global_concurrency_limit`.  This change should not affect current usage of the deployment resource.
 - The API docs reference a field called `denied_slots` doesn't return a response and breaks the provider.  For now we are not including but will revisit if it is part of the API read response in the future. 

# Tests
- Unit tests were added to test the lifecycle (create, read, update, delete) of the `global_concurrency_limit` resource and all have passed as expected.


internal/provider/resources/global_concurrency_limit.go
```diff
@@ -0,0 +1,344 @@
...
+		return
+	}
+
+	r.client = client
+}
+
+// Schema defines the schema for the resource.
+func (r *GlobalConcurrencyLimitResource) Schema(_ context.Context, _ resource.SchemaRequest, resp *resource.SchemaResponse) {
+	resp.Schema = schema.Schema{
+		Description: "The resource `global_concurrency_limit` represents a global concurrency limit. Global concurrency limits allow you to control how many tasks can run simultaneously across all workspaces.",
```
mitchnielsen, line 83: Let's link out to the docs again here

internal/provider/resources/global_concurrency_limit.go
```diff
@@ -0,0 +1,344 @@
...
+				Description: "The number of active slots.",
+				Default:     int64default.StaticInt64(0),
+				Validators: []validator.Int64{
+					int64validator.AtLeast(0),
+				},
+			},
+			"slot_decay_per_second": schema.Float64Attribute{
+				Optional:    true,
+				Computed:    true,
+				Description: "Slot Decay Per Second (number) or Slot Decay Per Second (null) (Slot Decay Per Second)",
```
mitchnielsen, line 145:
```suggestion
				Description: "Slot Decay Per Second (number or null)",
```

internal/provider/resources/global_concurrency_limit.go
```diff
@@ -0,0 +1,344 @@
...
+		return
+	}
+
+	err = client.Delete(ctx, state.ID.ValueString())
+	if err != nil {
+		resp.Diagnostics.Append(helpers.ResourceClientErrorDiagnostic("Global Concurrency Limit", "delete", err))
+
+		return
+	}
+}
```
mitchnielsen, line 227:
```suggestion
	resp.Diagnostics.Append(resp.State.Set(ctx, &state)...)
	if resp.Diagnostics.HasError() {
		return
	}
}
```

You'll need to update the state to reflect that the object is deleted

mitchnielsen (changes requested):
This is looking good, just a couple small comments

mitchnielsen (approved):
Nice work 🚀

PR #369, opened 2025-02-04
chore(tests): implement state checks for resources

mitchnielsen:
## Summary

Implements [state checks](https://developer.hashicorp.com/terraform/plugin/testing/acceptance-tests/state-checks) to replace the legacy `Checks` for `resources` (`datasources` to come after, already a big change).

This is mostly a drop-in replacement, but this new package gives us some extra helpers and allows us to check specific types instead of every value being a string.

Related to https://linear.app/prefect/issue/PLA-998/use-state-checks-in-tests


jamiezieziula (approved):

jimid27 (approved):
You walked me through some of this PR earlier and everything makes sense to me so 🚢 it

PR #371, opened 2025-02-06
chore(tests): standardize imports and their tests

mitchnielsen:
## Summary

Closes https://linear.app/prefect/issue/PLA-999/standardize-import-checks

- Introduces common `ImportStateBy{Name,ID}` functions to reuse among any resources that support importing by `id`/`name` and/or `workspace_id`
- Consistently uses `GetResourceWorkspaceImportStateID` instead of separate but very similar functions to do the same thing.
- Updates the import instructions to reflect the correct argument order.

## Notes

A few resources deviate from the `<identifier>,<workspace_id>` pattern:

```
$ rg 'name\/|handle\/' internal/provider/resources
internal/provider/resources/workspace.go:352:32:        if strings.HasPrefix(req.ID, "handle/") {
internal/provider/resources/workspace.go:353:41:                handle := strings.TrimPrefix(req.ID, "handle/")
internal/provider/resources/service_account_test.go:192:44:                             ImportStateIdPrefix:                  "name/",
internal/provider/resources/workspace_test.go:62:27:                            ImportStateIdPrefix: "handle/",
internal/provider/resources/service_account.go:508:32:  if strings.HasPrefix(req.ID, "name/") {
internal/provider/resources/service_account.go:509:39:          name := strings.TrimPrefix(req.ID, "name/")
internal/provider/resources/variable.go:504:4:// name/<variable_name>
internal/provider/resources/variable.go:505:4:// name/<variable_name>,<workspace_id>
internal/provider/resources/variable.go:522:44: if strings.HasPrefix(variableIdentifier, "name/") {
internal/provider/resources/variable.go:523:51:         name := strings.TrimPrefix(variableIdentifier, "name/")
```

So this includes workspaces, service accounts, and variables. These allow you to use a different prefix like `name/` or `handle/` to import the resource. The logic for these is fairly simple, so to keep this PR as small as possible I left those alone for now.

I also wanted to update the tests to import by ID alone, rather than ID + workspace ID. However, this didn't work because the test client doesn't have access to the `Config` resource to get the workspace value. Looks like this is related to https://github.com/hashicorp/terraform-plugin-testing/issues/11. I did build the provider and test against a real instance, though, and importing a resource by ID alone still worked.

### DIFF
--- a/docs/resources/automation.md
+++ b/docs/resources/automation.md
@@ -527,14 +527,9 @@ Optional:
 Import is supported using the following syntax:
 
 ```shell
-# prefect_automation resources can be imported by the Automation ID
+# prefect_automation resources can be imported by the automation_id
 terraform import prefect_automation.my_automation 00000000-0000-0000-0000-000000000000
-
-# Pass an optional, comma-separated value following the identifier
-# if you need to import a resource in a different workspace
-# from the one that your provider is configured with
-# NOTE: you must specify the workspace_id attribute in the addressed resource
 #
-# <automation_id>,<workspace_id>
+# or from a different workspace via automation_id,workspace_id
 terraform import prefect_automation.my_automation 00000000-0000-0000-0000-000000000000,11111111-1111-1111-1111-111111111111
 ```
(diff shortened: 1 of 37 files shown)


jamiezieziula (approved):
love this

PR #372, opened 2025-02-07
chore: add pull request template

mitchnielsen:
Adding a pull request template for new resources so we have a good reminder of all the boxes we need to check before adding a feature.

Related to https://linear.app/prefect/issue/PLA-983/cycle-12-catch-all


jamiezieziula (approved):
💯

PR #384, opened 2025-02-19
feat: add support for work queues (v2)

mitchnielsen:
### Summary

Closes https://linear.app/prefect/issue/PLA-193/add-support-for-work-queues

### Requirements

- [x] Title follows the format: `feat(<resource name>): support <resource name>`
- [x] Summary includes `Closes <issue>`, if available
- [x] Unit tests are added, when needed
- [x] Acceptance tests are added (including import tests, when needed)


docs/resources/work_queue.md
```diff
@@ -0,0 +1,69 @@
...
+- `id` (String) Work queue ID (UUID)
+- `updated` (String) Timestamp of when the resource was updated (RFC3339)
+
+## Import
+
+Import is supported using the following syntax:
+
+```shell
+# Prefect Work Queues can be imported using work_pool_name,work_queue_name,workspace_id
+terraform import prefect_work_queue.example kubernetes-work-pool,my-work-queue,00000000-0000-0000-0000-000000000000
```
mitchnielsen, line 68: I believe there's an API route to get work queues by ID, too - let me look into that.

mitchnielsen: Interesting - so we're actually hitting the `work pools` API route, which supports `Read work queue by name` but not `Read work queue (by ID)`.

Looks like we're doing this because we initialize the client with the name of the work pool, probably so we can support the `work_queues` datasource that lists all the queues on a given work pool.

This is fine, it just means it probably doesn't make much sense to add another function `GetByID` because it would hit an entirely different API (work queue API instead of the current work pool API) and would get a little messy.

I'm not opposed to this in the future, but since you can already import work queues by work pool name + work queue name, I think that's sufficient for now.

jimid27: This makes sense in my brain

internal/client/work_queues.go
```diff
@@ -0,0 +1,143 @@
...
+		body:         http.NoBody,
+		apiKey:       c.apiKey,
+		basicAuthKey: c.basicAuthKey,
+	}
+
+	resp, err := request(ctx, c.hc, cfg)
+	if err != nil {
+		return fmt.Errorf("failed to delete work queue: %w", err)
+	}
+	defer resp.Body.Close()
```
jimid27, line 140: TIL about `defer` nifty

jimid27 (approved):
This lgtm 🚢

PR #402, opened 2025-03-04
feat: add account_member resource

### LINKED ISSUE #78: Add support for Account Member
Note:
* we do not expose a create endpoint for `account_memberships` -- these are handled upon creation of the actors themselves (bots / users)
* However, we could still create a Terraform resource, but this would require that practitioners always import the resource via `terraform import prefect_account_membership`
* Another way to handle this could be that upon `Create` of the resource, we query for `account_memberships` based on the `email` attribute, and then hit the PATCH endpoint with the fetched membership ID.  `Update` wouldn't be a problem, as we could just use the membership ID in state

This issue will be deprioritized for the initial provider launch

mitchnielsen:
### Summary

Adds support for an `account_member` resource to manage account membership. This currently only allows changing the account role, as that's the only field supported in the `Update`/`Patch` operation.

Closes https://linear.app/prefect/issue/PLA-196/add-support-for-account-member

Closes https://github.com/PrefectHQ/terraform-provider-prefect/issues/78

#### General

#### Code-level changes


.github/workflows/acceptance-tests.yaml
```diff
@@ -42,7 +42,7 @@ jobs:
           go-version-file: 'go.mod'
       - uses: hashicorp/setup-terraform@v3
         with:
-          terraform_version: '1.5.*'
+          terraform_version: '1.11.*'
```
mitchnielsen, line 45: Looks like CI won't use this new version, even though viewing the workflow file says it should be 1.11, it keeps installing 1.5. Opened a separate PR so it's in main, assuming that'll fix it: https://github.com/PrefectHQ/terraform-provider-prefect/pull/404

jimid27 (approved):
nice this lgtm

PR #404, opened 2025-03-06
chore(acceptance tests): use Terraform 1.11

mitchnielsen:
### Summary

The 'removed' block is only supported in Terraform 1.7 and higher, but we were pinning to 1.5. We're currently using 1.11 in .mise.toml, so let's just match them.

Related to https://linear.app/prefect/issue/PLA-1193/cycle-14-catch-all

Related to https://github.com/PrefectHQ/terraform-provider-prefect/pull/402#discussion_r1982472715

#### General

### DIFF
--- a/.github/workflows/acceptance-tests.yaml
+++ b/.github/workflows/acceptance-tests.yaml
@@ -42,7 +42,7 @@ jobs:
           go-version-file: 'go.mod'
       - uses: hashicorp/setup-terraform@v3
         with:
-          terraform_version: '1.5.*'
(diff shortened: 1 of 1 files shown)


jimid27 (approved):
Should this be a version that we update through updatecli moving forward or is it ideal to pin this and update manually as we go?

mitchnielsen:
Good idea, we should do that. Opened https://linear.app/prefect/issue/PLA-1214/manage-terraform-version-upgrades-in-ci to track.

Issue #264, opened 2024-09-19
Add support for user API keys


mitchnielsen:
## Summary

Context: https://github.com/PrefectHQ/terraform-provider-prefect/issues/255#issuecomment-2361793641

> Will there be a Terraform Resource for Account API Keys developed?

> I know that there is one for ServiceAccount which in turn gives an API key with a permission set, but that requires to be on a paid version of Prefect. As a developer whom cant afford that at the current moment, I was wondering if you and your fellow incredible devs were thinking about creating that resource? For people whom are lower on the totem pole and are utilizing the lowest tier of Prefect, this Terraform Resource would truly help with credential rotation to keep an environment as secure as possible.

mitchnielsen:
@scondo-prof - to make sure we're talking about the same API: are you referring to [User API key](https://app.prefect.cloud/api/docs#tag/Users/operation/create_user_api_key_api_users__id__api_keys_post)? (`api/users/{id}/api_keys`)

scondo-prof:
@mitchnielsen Yes this is indeed what I am talking about!

Also let me add, I am very appreciative of your thorough communication!

mitchnielsen:
Great, we'll follow up here with any updates on progress.

And thanks, much appreciated! Reach out if you have any other questions/feedback.

PR #409, opened 2025-03-07
feat: add `user` resource

### LINKED ISSUE #410: Add support for User resource
## Summary

In the context of https://github.com/PrefectHQ/terraform-provider-prefect/issues/264, let's add a `user` resource to manage users with Terraform. We won't be able to create them, since there's no API endpoint for that, but we can import the resource and then manage most of the attributes on it from there.

This will also enable us to attach `user` resource IDs to `user_api_key` resources when we support them. We can also use `user` resources to add actor IDs to team membership resources.

mitchnielsen:
### Summary

Adds a `user` resource using the [users API](https://app.prefect.cloud/api/docs#tag/Users/operation/read_user_api_users__id__get).

I originally set out to add user API keys to this resource, roughly following the pattern we use in the `service_account` resource. But that ended up not feeling quite right because user API keys don't have a `Rotate` endpoint, so the logic ends up working a bit differently. I wasn't seeing a way to cleanly handle when to create or delete the token without really overloading the `user`'s `Update` method.

I see this working a bit better with a dedicated `user_api_key` resource. Adding the `user` resource is still helpful because there are some update-able fields on that resource that users may want to manage, and it also retrieves the actor ID which can be used for managing team membership. And then, of course, it would be used in the `user_api_key` resource by being able to provide the user's ID attribute in that resource.

Closes https://github.com/PrefectHQ/terraform-provider-prefect/issues/410

Closes https://linear.app/prefect/issue/PLA-1217/add-support-for-user-resource

Related to https://github.com/PrefectHQ/terraform-provider-prefect/issues/264

Related to https://linear.app/prefect/issue/PLA-346/add-support-for-api-keys

#### General

#### Code-level changes


internal/client/users.go
```diff
@@ -0,0 +1,89 @@
...
+	return &UsersClient{
+		hc:           c.hc,
+		apiKey:       c.apiKey,
+		basicAuthKey: c.basicAuthKey,
+		routePrefix:  fmt.Sprintf("%s/users", c.endpoint),
+	}, nil
+}
+
+// Read reads a user.
+func (c *UsersClient) Read(ctx context.Context, userID string) (*api.User, error) {
```
mitchnielsen, line 33: This is currently failing - I have a feeling you can't hit this endpoint if you aren't that user. I can hit my own user with my own API token, but can't hit Marvin's user.

mitchnielsen: If the token we use in GitHub CI was generated by someone (I'm guessing Eddie?) then we could potentially test against that account.

Alternatively, we could generate an API token as the Marvin user and use that in CI.

mitchnielsen: If this is all correct, we'll want to be clear that you can only use this to manage your personal user - not other users.

.github/workflows/acceptance-tests.yaml
```diff
@@ -52,3 +52,19 @@ jobs:
           PREFECT_API_URL: ${{ secrets.ACC_TEST_PREFECT_API_URL }}
           PREFECT_API_KEY: ${{ secrets.ACC_TEST_PREFECT_API_KEY }}
           PREFECT_CLOUD_ACCOUNT_ID: ${{ secrets.ACC_TEST_PREFECT_CLOUD_ACCOUNT_ID }}
+      - name: Run acceptance tests (user resource)
```
mitchnielsen, line 55: This won't actually run until merged to `main`, so I'll keep an eye out for any failures after this is merged.

internal/provider/resources/user.go
```diff
@@ -0,0 +1,243 @@
...
+		return
+	}
+
+	copyUserToModel(user, &plan)
+
+	resp.Diagnostics.Append(resp.State.Set(ctx, plan)...)
+}
+
+// Delete deletes a user.
+func (r *UserResource) Delete(ctx context.Context, req resource.DeleteRequest, resp *resource.DeleteResponse) {
```
mitchnielsen, line 207: Thinking out loud: should we instead _not_ implement Delete? I don't think an end user would ever truly want to delete a User if they remove the configuration from Terraform. Deleting users is, as far as I know, a more involved process that also includes removing API keys, memberships, workspaces, personal accounts, etc.

If we did go that way, we could also remove the lifecycle deletion protection on the acceptance tests.

internal/client/users.go
mitchnielsen: The Users API does indeed check to ensure "IsSelf", so users can only retrieve/manage information for their own user. Because of this, I created an API key as the Marvin robot user we have in the `github-ci-tests` workspace and added it as a GitHub secret. I also added a second acceptance testing CI step that tests only the `user` resource, providing the different API key and test name.

.github/workflows/acceptance-tests.yaml
mitchnielsen: I did test it locally by passing in the values myself and it passed.

internal/provider/resources/user.go
parkedwards: im actually not sure of the answer - im assuming that a valid DELETE endpoint exists

however, i'd advocate for omitting DELETE in tandem with CREATE - they should mirror each other in terms of what we allow the provider to manage.  in our case, we omit CREATE, so I would say we may want to omit DELETE to avoid the issue you raised

internal/provider/resources/user_test.go
```diff
@@ -0,0 +1,106 @@
...
+func fixtureAccUserResource(userID string) string {
+	return fmt.Sprintf(`
+resource "prefect_user" "test" {
+  id = "%s"
+
+  # This is here as a safeguard to prevent the user from being
+  # destroyed when the test is finished. This is primarily handled by the
+  # 'fixtureAccUserResourceUnmanage' variable.
+  lifecycle {
+    prevent_destroy = true
```
parkedwards, line 21: 👍

parkedwards (approved):
LGTM!

internal/provider/resources/user.go
mitchnielsen: That sounds good, updated in 8bae2dd.

PR #419, opened 2025-03-14
chore(CI): fix user tests

mitchnielsen:
### Summary

In [this workflow run](https://github.com/PrefectHQ/terraform-provider-prefect/actions/runs/13847776057/job/38749553993?pr=418), I noticed that the user tests don't seem to be called - instead it looks like just the entire test suite is running.

Related to https://linear.app/prefect/issue/PLA-880/cycle-10-catch-all

### Testing

See that now the [job](https://github.com/PrefectHQ/terraform-provider-prefect/actions/runs/13865165235/job/38802576945?pr=419#step:8:25) only runs 2 tests in the `resources` package.

Also:
- [Intentional test failure](https://github.com/PrefectHQ/terraform-provider-prefect/actions/runs/13865941575/job/38804933912?pr=419) to confirm the tests are indeed trying to run and not being skipped
- [Successful test run](https://github.com/PrefectHQ/terraform-provider-prefect/actions/runs/13866193068/job/38805711460?pr=419) after removing the intentional bug

#### General

#### Code-level changes

- [ ] Documentation is added (generated by `make docs` from source code)
- [ ] For resources, the following are added:


Makefile
```diff
@@ -42,13 +43,14 @@ lint:
 .PHONY: lint
 
 test:
-	gotestsum --max-fails=50 ./...
+	gotestsum --max-fails=50 ./... -run "$(TESTS)"
```
mitchnielsen, line 46: This is the change that makes sure only the specified tests are passed. We didn't need this previously because we always ran all tests.

Makefile
```diff
@@ -42,13 +43,14 @@ lint:
...
+
 .PHONY: test
 
 # NOTE: Acceptance Tests create real infrastructure
 # against a dedicated testing account
 testacc:
-	TF_ACC=1 make test
+	TF_ACC=1 TESTS=$(TESTS) make test
```
mitchnielsen, line 53: Accompanying change to pass the specified tests along.

internal/provider/resources/user_test.go
```diff
@@ -36,8 +38,8 @@ removed {
 //  1. Provide a value for `api_key` from an API key generated as a user,
 //     not a service account.
 //  2. Provide the user ID as an environment variable: `ACC_TEST_USER_RESOURCE_ID`
-func SkipIfUserResource() (bool, error) {
-	if os.Getenv("ACC_TEST_USER_RESOURCE") == "yes" {
+func SkipUnlessUserResource() (bool, error) {
+	if _, set := os.LookupEnv(envUserResourceID); set {
```
mitchnielsen, line 42: Simplifies the logic here a bit, just checking if the user resource ID is set at all. This way we don't need yet another env var set to "yes" to run the tests.

.github/workflows/acceptance-tests.yaml
```diff
@@ -59,12 +62,15 @@ jobs:
...
           # This is the user ID retrieved by logging in as the Marvin user
           # and copying the ID on the "Profile" page.
           ACC_TEST_USER_RESOURCE_ID: "48750018-b4c4-4484-8fbf-b61baf3926b5"
           # This ensures only the User resource tests run, instead of running
           # all the tests again because that was completed in the last step.
-          TESTS: "TestAccResource_user"
+          # Regex ensures that the following tests are included:
+          # - TestAccResource_user
+          # - TestAccResourcE_user_api_key
+          TESTS: "TestAccResource_user*"
```
parkedwards, line 76: good catch - i thought it was already doing this pattern match, but i guess it was running all the tests

parkedwards (approved):

PR #552, opened 2025-07-15
feat(block resource): add write-only data attribute

### LINKED ISSUE #549: Feature Request: Write-only data for prefect_block resource
Could you please extend the `prefect_block` resource to support a [write-only attribute](https://developer.hashicorp.com/terraform/plugin/sdkv2/resources/write-only-arguments) for `data`? Alternatively, an additional resource type could also be introduced.

I tried to create a webhook `prefect_block` retrieving a required API key from an existing secret in AWS. To avoid credentials being stored in the Terraform state, I prefer using `ephemeral` blocks. However, my attempt failed with the error message `Ephemeral values are not valid for "data", because it is not a write-only attribute and must be persisted to state.`

Here's the code I tested:

```terraform
ephemeral "aws_secretsmanager_secret_version" "api_key" {
  secret_id = "api-key"
}

resource "prefect_block" "dummy_webhook" {
  name      = "dummy-webhook"
  type_slug = "webhook"
  data = jsonencode({
    url: "..."
    headers: {
      "API-KEY": ephemeral.aws_secretsmanager_secret_version.api_key.secret_string
    }
  })
}
```

Maybe this could be implemented similar to the [AWS secretsmanager_secret_version resource](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/secretsmanager_secret_version):

```terraform
ephemeral "aws_secretsmanager_secret_version" "api_key" {
  secret_id = "api-key"
}

resource "prefect_block" "dummy_webhook" {
  name      = "dummy-webhook"
  type_slug = "webhook"
  data_wo = jsonencode({
    url: "..."
    headers: {
      "API-KEY": ephemeral.aws_secretsmanager_secret_version.api_key.secret_string
    }
  })
  data_wo_version = "1"
}
```

mitchnielsen:
### Summary

Adds an attribute called `data_wo` to the block resource, providing a write-only attribute for security purposes.

References:
- https://developer.hashicorp.com/terraform/plugin/framework/resources/write-only-arguments
- https://developer.hashicorp.com/terraform/language/resources/ephemeral/write-only

Closes https://github.com/PrefectHQ/terraform-provider-prefect/issues/549

See also https://linear.app/prefect/issue/PLA-1028/implement-ephemeral-resources

### Testing

First, see the acceptance tests added. They confirm that:

- Write-only data attribute changes result in a `null` value in the state (as expected)
- Write-only data attribute changes trigger an update
- Write-only data attributes staying the same does _not_ trigger an update

Additionally, testing by hand:

```terraform
resource "prefect_block" "dummy_secret" {
  name      = "dummy-secret"
  type_slug = "secret"

  data_wo         = jsonencode({ "value" = "my-secret-value" })
  data_wo_version = 1
}
```

And confirming the result is null:

```
$ jq '.resources[0].instances[0].attributes.data_wo' terraform.tfstate
null

$ tf state show prefect_block.dummy_secret
# prefect_block.dummy_secret:
resource "prefect_block" "dummy_secret" {
    created         = "2025-07-15T21:46:40Z"
    data_wo         = (write-only attribute)
    data_wo_version = 1
    id              = "80bb8cd5-b1c4-4bc7-911d-94580073b8f8"
    name            = "dummy-secret"
    type_slug       = "secret"
    updated         = "2025-07-15T21:46:53Z"
}
```

Can confirm the value with Python:

```python
from prefect.blocks.system import Secret

secret_block = Secret.load("dummy-secret")
print(secret_block.get())
```

```
$ python test.py
my-secret-value
```

You can then:
- Update the value, apply, and see nothing changes
- Update the version too, apply, and see the value change

#### General

#### Code-level changes

### DIFF
--- a/examples/resources/prefect_block/resource.tf
+++ b/examples/resources/prefect_block/resource.tf
@@ -13,7 +13,7 @@ resource "prefect_block" "secret" {
 }
 # example:
 # you can also use a write-only attribute for the data field
-resource "prefect_block" "secret_write_onlyh" {
+resource "prefect_block" "secret_write_only" {
   name = "foo-write-only"
 
   # prefect block type ls
--- a/examples/resources/prefect_block/resource.tf
+++ b/examples/resources/prefect_block/resource.tf
@@ -11,6 +11,20 @@ resource "prefect_block" "secret" {
   # set the workpace_id attribute on the provider OR the resource
   workspace_id = "<workspace UUID>"
 }
+# example:
+# you can also use a write-only attribute for the data field
+resource "prefect_block" "secret_write_onlyh" {
+  name = "foo-write-only"
+
+  # prefect block type ls
+  type_slug = "aws-credentials"
+
+  # prefect block type inspect aws-credentials
+  data_wo = file("./aws-credentials.json")
+
+  # provide the version to control when to update the block data
+  data_wo_version = 1
+}
 
 # example:
 # you can also use file() to import a JSON file
--- a/examples/resources/prefect_block/resource.tf
+++ b/examples/resources/prefect_block/resource.tf
@@ -20,7 +20,9 @@ resource "prefect_block" "secret_write_onlyh" {
   type_slug = "aws-credentials"
 
   # prefect block type inspect aws-credentials
-  data_wo = file("./aws-credentials.json")
+  data_wo = jsonencode({
+    "value" = "bar"
+  })
 
   # provide the version to control when to update the block data
   data_wo_version = 1
--- a/internal/provider/resources/block.go
+++ b/internal/provider/resources/block.go
@@ -299,7 +299,7 @@ func (r *BlockResource) Create(ctx context.Context, req resource.CreateRequest, 
 		return
 	}
 
-	if dataWO != nil {
+	if len(dataWO) != 0 {
 		data = dataWO
 	}
 
--- a/internal/provider/resources/block.go
+++ b/internal/provider/resources/block.go
@@ -7,11 +7,15 @@ import (
 	"github.com/avast/retry-go/v4"
 	"github.com/google/uuid"
 	"github.com/hashicorp/terraform-plugin-framework-jsontypes/jsontypes"
+	"github.com/hashicorp/terraform-plugin-framework-validators/int32validator"
+	"github.com/hashicorp/terraform-plugin-framework-validators/stringvalidator"
 	"github.com/hashicorp/terraform-plugin-framework/diag"
+	"github.com/hashicorp/terraform-plugin-framework/path"
 	"github.com/hashicorp/terraform-plugin-framework/resource"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema/planmodifier"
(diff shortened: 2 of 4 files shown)


jamiezieziula (approved):

examples/resources/prefect_block/resource.tf
```diff
@@ -11,6 +11,22 @@ resource "prefect_block" "secret" {
   # set the workpace_id attribute on the provider OR the resource
   workspace_id = "<workspace UUID>"
 }
+# example:
+# you can also use a write-only attribute for the data field
+resource "prefect_block" "secret_write_onlyh" {
```
parkedwards, line 16:
```suggestion
resource "prefect_block" "secret_write_only" {
```

internal/provider/resources/block.go
```diff
@@ -117,10 +123,34 @@ func (r *BlockResource) Schema(_ context.Context, _ resource.SchemaRequest, resp
 				},
 			},
 			"data": schema.StringAttribute{
-				Required:    true,
+				Optional:    true,
+				Sensitive:   true,
+				CustomType:  jsontypes.NormalizedType{},
+				Description: "The user-inputted Block payload, as a JSON string. Use `jsonencode` on the provided value to satisfy the underlying JSON type. The value's schema will depend on the selected `type` slug. Use `prefect block type inspect <slug>` to view the data schema for a given Block type.",
+				Validators: []validator.String{
+					stringvalidator.ConflictsWith(
```
parkedwards, line 131: nice, so only `data` or `data_wo` can be set?

internal/provider/resources/block.go
```diff
@@ -383,12 +447,22 @@ func (r *BlockResource) Update(ctx context.Context, req resource.UpdateRequest,
...
 
+	// If the data_wo_version is different, we need to update the data payload
+	// using the content of the data_wo attribute.
+	if !plan.DataWOVersion.Equal(state.DataWOVersion) {
+		data, diags = helpers.UnmarshalOptional(config.DataWO)
+		resp.Diagnostics.Append(diags...)
+		if resp.Diagnostics.HasError() {
+			return
+		}
+	}
```
parkedwards, line 464: im guessing this is needed because there's no way to actually tell if `data_wo` has changed by the value alone, because we dont store it

parkedwards (approved):
dude this is nice! pretty elegant way to handle diff changes via `data_wo_version`. would the alternative be to always write _**something**_ on every plan/apply?

mitchnielsen:
> dude this is nice! pretty elegant way to handle diff changes via `data_wo_version`. would the alternative be to always write _**something**_ on every plan/apply?

That's my understanding yup. The `version` approach is something I found in the [best practices section of the docs](https://developer.hashicorp.com/terraform/plugin/framework/resources/write-only-arguments#best-practices) and the [Google provider](https://registry.terraform.io/providers/hashicorp/google-beta/latest/docs/guides/using_write_only_attributes). Seems to work well to avoid having a non-empty plan every time

internal/provider/resources/block.go
mitchnielsen: Yep exactly, there isn't an explicit example of how to handle this in the docs but this seems to be roughly what other providers do.

PR #576, opened 2025-09-09
feat: add profile file auth

### LINKED ISSUE #575: Use local profiles for authentication
As a Prefect administrator, I'd like to be able to use my local profiles in my home directory for authentication.

bdalpe:
### Summary

Adds the ability to read from the user's local Prefect profiles file and use the contents as a source of authentication. This method is tried after the local provider config and explicitly set environment variables.

Closes #575 

#### General

#### Code-level changes

- [x] Documentation is added (generated by `mise run docs` from source code)


internal/provider/provider.go
```diff
@@ -146,12 +156,33 @@ func (p *PrefectProvider) Configure(ctx context.Context, req provider.ConfigureR
...
+	}
+
 	// Extract endpoint from configuration or environment variable.
 	var endpoint string
 	if !config.Endpoint.IsNull() {
 		endpoint = config.Endpoint.ValueString()
 	} else if apiURLEnvVar, ok := os.LookupEnv(envAPIURL); ok {
 		endpoint = apiURLEnvVar
+	} else if !profileAuth.Endpoint.IsNull() {
+		endpoint = profileAuth.Endpoint.ValueString()
```
mitchnielsen, line 185: Now that we have two implicit options for configuration (env vars and profiles file), what do you think of adding a diagnostic somewhere in the output for the user to confirm which method was used? I'm picturing a situation where they have env vars set but also have a profiles file, and want to be sure that configuration is being pulled from the env vars (given that they're higher precedence).

mitchnielsen:
@bdalpe for sure, but just to clarify - those testing notes were showing that it worked successfully ✅ 

```terraform
terraform {
  required_providers {
    prefect = {
      source = "prefecthq/prefect"
    }
  }
}

provider "prefect" { }

resource "prefect_variable" "variable" {
  name  = "mitch_test_profiles"
  value = "helloworld"
}
```

internal/provider/provider.go
mitchnielsen: I think it'd be good, even if it's only a debug level log - especially since this is implicitly reading configuration without any direct user guidance

mitchnielsen:
Acceptance Tests OSS failing is unrelated, see https://linear.app/prefect/issue/PLA-2093/terraform-provider-fails-with-inconsistent-result-after-prefect-oss.

mitchnielsen (approved):

PR #652, opened 2026-03-13
fix(resources): reset work pool access to wildcard on deletion

Skoczek135:
### Summary

- Fix `Delete` on `prefect_work_pool_access` to reset actor IDs to wildcard (`"*"`) instead of empty slices. Empty slices resulted in a restrictive state where no one had access to the work pool after resource deletion, rather than restoring the default open access.
- Add acceptance test for the delete path using a two-step approach: first step creates the access control, second step removes only the `prefect_work_pool_access` resource (keeping the work pool alive) and verifies that access control is reset to wildcard defaults.

#### General

- [x] Title follows the [conventional commits](https://www.conventionalcommits.org/) format

#### Code-level changes

IMG_3627

### DIFF
--- a/internal/provider/resources/work_pool_access.go
+++ b/internal/provider/resources/work_pool_access.go
@@ -347,7 +347,6 @@ func (r *WorkPoolAccessResource) Delete(ctx context.Context, req resource.Delete
 		return
 	}
 
-
 	payload := api.WorkPoolAccessSet{}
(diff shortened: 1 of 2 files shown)


mitchnielsen (changes requested):
Looks great overall. Left two small suggestions that should fix the linting error.

mitchnielsen (approved):

PR #677, opened 2026-05-06
fix(resources): preserve user rrule round-trip for deployment_schedule

mitchnielsen:
### Summary

Prefect OSS now normalizes rrule values by prepending an explicit `DTSTART:20200101T000000\n` line on write, so the GET response no longer matches what the provider sent. Terraform sees plan vs. state divergence and aborts with `Provider produced inconsistent result after apply`, which has been failing step 4 of `TestAccResource_deployment_schedule` on every PR (e.g. [run 25409601316](https://github.com/PrefectHQ/terraform-provider-prefect/actions/runs/25409601316)).

This strips a leading `DTSTART...\n` line from the server response when the prior local value didn't itself have one. Users who supply their own DTSTART get the server's value verbatim — no permanent diff in either direction.

Verified end-to-end: `./scripts/testacc-oss TestAccResource_deployment_schedule` passes all 5 steps against `prefecthq/prefect:3-latest`. The interval and cron schedule kinds were already passing, so the same class of issue isn't lurking there.

Closes [PLA-2677](https://linear.app/prefect/issue/PLA-2677/prefect-deployment-schedule-rrule-round-trip-fails-after-server)

#### General

#### Code-level changes

<details>
<summary>Session context</summary>

Considered three approaches from the issue:

- **(a) Accept the server-normalized form into state.** Cleanest at the seam, but flips observable state for every existing rrule user on upgrade.
- **(b) Strip `DTSTART:...\n` from server responses.** Simple, but if applied unconditionally would clobber user-supplied DTSTARTs.
- **(c) Always send rrule with explicit DTSTART.** Mutates user input and locks the provider to whatever anchor the server happens to use today.

Landed on (b) with a guard: compare the server response against the prior local value (the plan during Create, prior state during Read/Update) and only strip the `DTSTART:` line when the local value didn't have one. That keeps both shapes round-trippable.

The helper is `normalizeRRuleForState(serverRRule, priorRRule)` plus a small `hasDTStartPrefix` check that handles both `DTSTART:` and `DTSTART;TZID=...:` forms, case-insensitive. White-box unit tests live in `deployment_schedule_internal_test.go`.

</details>

### DIFF
--- a/internal/provider/resources/deployment_schedule.go
+++ b/internal/provider/resources/deployment_schedule.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"strings"
 
 	"github.com/hashicorp/terraform-plugin-framework-jsontypes/jsontypes"
 	"github.com/hashicorp/terraform-plugin-framework/diag"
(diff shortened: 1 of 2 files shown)


parkedwards (approved):
nice tests

PR #680, opened 2026-06-08
feat: support running acceptance tests against a customer-managed instance

mitchnielsen:
This makes the provider and its acceptance test suite work against a customer-managed (CM) Prefect instance, in addition to Prefect Cloud and OSS.

Related to https://linear.app/prefect/issue/CUS-651

Secrets added via https://github.com/PrefectHQ/platform/pull/11336.

The primary change to functionality, outside of tests: reserve the deployment-schedule `slug` when the server response omits it. CM persists the slug but its `DeploymentScheduleResponse` schema doesn't echo it back, which tripped "inconsistent result after apply". This mirrors the existing rrule/parameters round-trip handling.

The rest is test-side work to make the suite portable:

<details>
<summary>Click to expand test-side changes</summary>

- A `CM` test context (`TEST_CONTEXT=CM`) mirroring the existing OSS one, with `SkipTestsIfCM` / `SkipFuncCM` / `SkipFuncOSSOrCM` / `SkipTestsIfOSSOrCM` helpers.
- Skips for features that are Cloud-only on CM (verified against the customer-managed server source): SSO/domain names, metric-trigger automations, send-email / pause-schedule automation actions, account `managed_execution` settings, deployment global concurrency limits, typed (non-string) variable values, and resource SLAs.
- Portability fixes that keep coverage on all environments: assert a minimum account-role permission count instead of an exact one, derive the webhook endpoint host from the client instead of hardcoding it, and read the pre-existing account `name`/`handle`/`link` from env vars.
- A `scripts/testacc-cm` runner and the `string` block type in fixtures (the GitHub/S3 block types aren't registered on CM by default).

</details>

There's also a new provider guide (`docs/guides/customer-managed.md`) describing how to point the provider at a Customer-Managed instance and which Cloud features aren't supported there. It links out to https://docs-customer-managed.prefect.io/ as the authoritative source.


scripts/testacc-dev
mitchnielsen, line 1: Looks like the vault item in 1P was modified a bit so this matches those changes.

jimid27 (approved):
this is awesome and lgtm thanks

